PR #11, opened 2019-04-11
Feat/uikit button docs

ciaranschutte:
~~Adds `storybook/readme` with sample usage.~~
Using storybook/info-addon to support storybook source code


client/uikit/Button/readme.md
```diff
@@ -0,0 +1,27 @@
...
+import Button from "uikit/button";
+```
+
+### Usage
+
+Basic
+
+```jsx
+<Button>My button text</Button>
+```
```
hlminh2000: Would it make sense to somehow tie this section to the story source? I'm thinking the story source should serve as the sample usage, to lower the cost of maintaining these documentations :) If such can be achieved with other tools, this section can be optional, mainly for pointing out context-specific stuff such as hidden dependencies / specific expected values for props, etc... Main concern being if the cost of maintenance is too high, doc will get out of sync from actual usage and becomes misleading. Especially in the early stage these components will likely change a lot, so the lower the cost of maintenance the better :)

The downside the storybook-specific code in the story can make the doc a bit noisy. I think the trade off is worth it in the early stage :)

What do you think?

client/uikit/Button/readme.md
```diff
@@ -0,0 +1,27 @@
+### Story
+
+<!-- STORY -->
+
+### Description
+
+A basic button
+
+<!-- PROPS -->
```
hlminh2000: more a question, I'm guessing this traverses the component tree and include proptypes of all the components used in the story? :)

client/uikit/Button/stories.js
```diff
@@ -1,10 +1,10 @@
...
 import React from "react";
 import Button from ".";
+import README from "./readme.md";
 
-const ButtonStories = storiesOf(`${__dirname}`, module).add(
-  "A basic button",
-  () => <Button>Cognito Argo Sum</Button>
-);
+const ButtonStories = storiesOf(`${__dirname}`, module)
+  .addParameters({ readme: { content: `${README}`, sidebar: `${README}` } })
```
hlminh2000: anyway to automate this / would it make sense to? :)

client/uikit/Button/readme.md
ciaranschutte: yup!
multiple props

client/uikit/Button/readme.md
ciaranschutte: Yeah that makes a lot of sense. Earlier on I was confusing source of the component vs source of the storybook. I'm actually gonna add back in info-addon instead of storybook-readme then, it's the only plugin i've come across that handles adding the source well....even if it is a little more ugly :p

client/uikit/Button/stories.js
ciaranschutte: Doesn't make much to because now the added docs will be referring to one specific story.
Most of the time not much to add.We can have the component readme in a little story on it's own.

Now looks like: 
``` javascript
const ButtonStories = storiesOf(`${__dirname}`, module)
  .addParameters({ info: 'Docs referring to this one specific story'})
```

ciaranschutte:
@hlminh2000  using storybook/info-addon so we get the nice source code view

PR #13, opened 2019-04-11
Feat/add uikit root readme

ciaranschutte:


client/uikit/readme.md
```diff
@@ -0,0 +1,21 @@
...
+- components should meet a level of accessibilty
+
+## Structure
+
+Each component folder should have:
+
+- source
+- tests (for now some basic unit tests)
+- a Storybook story
+- a readme file documenting important points/decisions
```
hlminh2000: could make sense to have a template file sitting in a sample story? :) something that can be copy-pasted into new components :P

ciaranschutte: good idea!

ciaranschutte:
Changed folder structure to use generator instead
i.e `npm run create-component Input` will create skeleton folder + files

PR #22, opened 2019-04-25
Feat/webpack absolute paths

ciaranschutte:
fixes errors with absolute import
fixes errors with uikit .babelrc being at root level, moved it into uikit


ciaranschutte:
Merged in different branch

PR #23, opened 2019-04-26
Ego auth test

ciaranschutte:
- adds temporary (using a basic anchor for now) Google login button to authenticate
- `fetch` call to get Ego token
- adds injection of globals using `dotenv` based on Nextjs example
  - `create-react-app` global injection doesn't work because SSR
  - benefit of using `dotenv` is we can keep using out `.env` file and load a lot of vars, instead of duplicating in the `next.config.js` file

Functionality not included in this PR:
- validate JWT
- store user
- error handling


client/package.json
```diff
@@ -20,12 +20,15 @@
...
     "babel-polyfill": "^6.26.0",
     "dotenv": "^7.0.0",
+    "dotenv-webpack": "^1.7.0",
     "emotion-theming": "^10.0.10",
     "express": "^4.16.4",
     "graphql": "^14.2.1",
     "graphql-import": "^0.7.1",
     "graphql-tag": "^2.10.1",
     "isomorphic-fetch": "^2.2.1",
+    "isomorphic-unfetch": "^3.0.0",
```
hlminh2000: there's already `isomorphic-fetch`, any reason we're preferring `isomorphic-unfetch`? :)

client/common/injectGlobals.js
```diff
@@ -0,0 +1,2 @@
+export const egoApiRoot = process.env.EGO_API_ROOT;
+export const egoClientId = process.env.EGO_CLIENT_ID;
```
hlminh2000: let's update the `client/.env.schema` file to include these :)

client/next.config.js
```diff
@@ -1,11 +1,19 @@
...
 const path = require("path");
+const Dotenv = require("dotenv-webpack");
 
 module.exports = {
   webpack: config => {
     // Fixes npm packages that depend on `fs` module
     config.node = {
       fs: "empty"
     };
+    config.resolve.modules.push(path.resolve("./"));
```
hlminh2000: just to understand :) is this for absolute import?

ciaranschutte: yep!

client/package.json
ciaranschutte: nope, just missed it :)

PR #25, opened 2019-04-29
Routing setup

hlminh2000:
server-side routing setup for the pages
- Ego login redirect stores token in cookie to allow server-side authentication on first loads
- Login flow redirects user to target route after successful logins
- Setup for each route to handle custom authorization logic
- Common view for unauthorized access


ciaranschutte:
I know it's a WIP so not sure if you wanted me to look at this yet.
One typo and one question :) 
Typo - there is a double `NODE_PATH` in `.env.schema`

Question is, should we be doing the auth checks at the _app page level so we're not duplicating it on a per page basis?

hlminh2000:
> @hlminh2000 I know it's a WIP so not sure if you wanted me to look at this yet.
> One typo and one question :)
> Typo - there is a double `NODE_PATH` in `.env.schema`
> 
> Question is, should we be doing the auth checks at the _app page level so we're not duplicating it on a per page basis? Something like https://github.com/bgold0/nextjs-auth-skeleton/blob/master/pages/_app.js

that makes a lot of sense :) Our flow might be a little different, but it does make sense to use the `_app.js` for auth check. Thanks man for the suggestion!

PR #26, opened 2019-05-01
Feat/e2e test proto

ciaranschutte:
E2E protoype

- adds Nightwatch testing (uses selenium under hood)
  -- easier to read / more convenience than selenium-webdriver
  -- built in assertion lib and test runner
  -- plays nicely with browserstack

- adds config for both a local running dev e2e testing and Browserstack local testing

- adds script to pick testing


client/package.json
```diff
@@ -6,9 +6,12 @@
...
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "create-component": "plop --plopfile ./uikit/plopfile.js",
     "flow": "flow",
-    "flow-typed": "flow-typed"
+    "flow-typed": "flow-typed",
+    "create-component": "plop --plopfile ./uikit/plopfile.js",
+    "e2e": "./e2e/scripts/interactive.bs.js",
+    "e2e-dev": "nightwatch -c e2e/conf/dev.js; exit 0",
+    "e2e-local": "./e2e/scripts/local.runner.js -c e2e/conf/local.js; exit 0"
```
ciaranschutte: leaving `e2e-dev` and `e2e-local` in for now in case `e2e` decides to break during demo

client/uikit/SocialLoginButtons/GoogleLogin/index.js
```diff
@@ -10,7 +10,9 @@ const StyledLink = styled("a")`
 `;
 
 const GoogleLogin = ({ link }) => (
-  <StyledLink href={link}>Log in with the Google</StyledLink>
+  <StyledLink id="google-log-in" href={link}>
```
hlminh2000: not sure how I feel about IDs in the uikit components :/ probably prefer `classname` but also make sure to handle `className` props in that case :)
Generally I think ids should only be set from application code. What do you think? :)

client/e2e/readme.md
```diff
@@ -0,0 +1,77 @@
...
+
+Use Nightwatch (http://nightwatchjs.org) to write tests.
+All tests located in `tests` folder
+Tests can be logically grouped into folders
+
+## Recording Tests
+
+At the time of writing Selenium IDE doesn't have export for scripts
+
+This recorder gets use half way there: https://github.com/vvscode/js--nightwatch-recorder
```
hlminh2000: `"gets use half way there"` typo? :P

PR #48, opened 2019-05-22
Feat/#48 inputs

ciaranschutte:


ciaranschutte:
@hlminh2000 working on select input in seperate PR

client/uikit/Input/index.js
```diff
@@ -0,0 +1,123 @@
...
+}) => {
+  const [isExpanded, setExpanded] = useState(false);
+
+  return (
+    <div>
+      <StyledInputWrapper
+        onClick={() => setExpanded(!isExpanded)}
+        disabled={disabled}
+        size={size}
+        inputState={activeState}
```
hlminh2000: `activeState` is not defined

kcullion:
@ciaranschutte some feedback:
1. As a default, can you remove the "Start typing here..." text? Sorry, I put that in my design but I didn't mean to leave it in as the input default
2. The disabled input can still be clicked on, activated and typed in. Shouldn't be able to type in it or activate it
3. Fonts aren't rendering correctly
4. Error: the input border should be red for every state
5.  The small and large sizes aren't rendering correctly - when toggle between them in the "knobs" section there is no difference to them. Also, the large input should be the default

hlminh2000:
@kcullion, `"Start typing here.."` is a configurable thing under the `placeholder` prop (you can type something else in the `Placeholder` input on the side and see it change)

@ciaranschutte, 
- regarding Kim's point about the search input, I think we can have a `SearchInput` component that just sets the defaults implementation for search boxes. Alternatively, we can have a `variant` field that defaults to `default` and `search` where the props defaults to the search configuration
- for the typograpy stuff, try using the `Typography` component that's been added ;)

kcullion:
Thanks @hlminh2000  - I just mean the typical input wouldn't have any placeholder text. So maybe it's easier to start with that state and allow the developer to type in placeholder texts if required

hlminh2000:
The Select component still needs some more work, but the input is looking good so I'm happy to move this forward and continue the work in later PRs

PR #56, opened 2019-05-22
Feat/#33 checkbox radio

ciaranschutte:


client/uikit/Radio/stories.js
```diff
@@ -0,0 +1,20 @@
...
+  const disabled = boolean("disabled", false);
+
+  return {
+    checked,
+    disabled
+  };
+};
+
+const RadioStories = storiesOf(`${__dirname}`, module).add("Basic", () => (
+  <Radio {...createKnobs()}>My Selection</Radio>
```
hlminh2000: maybe a couple more props added just for demo purpose, particularly the `onChange` would be nice to have, using storybook's `action` :)

ciaranschutte: @hlminh2000 just to log the action? It's a controlled component so the checked state is controlled by the storybook-knob or I could wrap these stories in a "form state" to pass the checked. thoughts?

hlminh2000: so I would just handle the "checked" state in the story state, it'll reflect in the doc anyway

hlminh2000: We can do something like, so both the event is visible and the state is flipped :)
```
checked={checked}
onChange={e => { 
  setChecked(true); 
  action("clicked")(e);
}}
```
I think it's a more intuitive thing for Kim to interact with

PR #64, opened 2019-05-24
User and Program schemas, fetching from Ego and Pierogi

joneubank:
Pierogi is just ProgramService with a better name.


server/services/programService/index.js
```diff
@@ -0,0 +1,73 @@
+/*
+ * This file dynamically generates a gRPC client from Ego.proto.
+ * The content of Ego.proto is copied directly from:
```
hlminh2000: comment seems to be cut off? :)

server/services/ego/index.js
```diff
@@ -0,0 +1,53 @@
+/*
+ * This file dynamically generates a gRPC client from Ego.proto.
+ * The content of Ego.proto is copied directly from:
```
hlminh2000: cut off comment?

hlminh2000:
btw, holding off on merging this until we get the upstream deploys figured out

PR #67, opened 2019-05-28
✨ sets up table component

hlminh2000:


client/uikit/Table/index.js
```diff
@@ -0,0 +1,196 @@
...
+
+import reactTableDefaultStyle from './reactTableDefaultStyle';
+import Typography from '../Typography';
+import ascending from '../assets/table/ascending.svg';
+import descending from '../assets/table/descending.svg';
+import unsorted from '../assets/table/unsorted.svg';
+import TablePagination from './TablePagination';
+import { useTheme } from '../ThemeProvider';
+
+const StyledTable = styled(ReactTable)`
```
joneubank: Do you want to move this to a styledComponents.js file like with other components?

client/uikit/Table/index.js
```diff
@@ -0,0 +1,196 @@
...
+export default Table;
+
+export const SelectTable = props => {
+  const { isSelected, data, keyField } = props;
+  const selectedIds = data.map(data => data[keyField]).filter(isSelected);
+  const Component = selectTable(Table);
+  return <Component {...props} isSelectTable primaryKey={keyField} selectedIds={selectedIds} />;
+};
+SelectTable.propTypes = {
+  ...SelectTable.propTypes,
```
joneubank: Should this be ReactTable.propTypes ? This is the only place SelectTable propTypes are defined so as written this looks unnecessary.

client/uikit/theme/defaultTheme/table.js
```diff
@@ -0,0 +1,7 @@
+import colors from "./colors";
+
+export default {
+  th: {
```
joneubank: Should this be a human readable name, not the html tag name?

joneubank:
Looks awesome. A couple small things to consider but nothing blocking.

client/uikit/Table/index.js
hlminh2000: great catch!

PR #88, opened 2019-06-05
Update page title bar

d8660091:


d8660091:
Close because of it's included in https://github.com/icgc-argo/argo-platform/pull/92

PR #104, opened 2019-06-10
Add create program page

d8660091:


client/pages/login.js
```diff
@@ -37,7 +37,7 @@ export default createPage({
     <div>
       <GoogleLogin
         id="google-login"
-        link={urlJoin(EGO_API_ROOT, `api/oauth/login/google?client_id=${EGO_CLIENT_ID}`)}
```
hlminh2000: no need to change this, it'd be a breaking change for deployment. Just put the ego domain in your local for development? :)

client/pages/create-program.js
```diff
@@ -0,0 +1,242 @@
...
+        </Col>
+      </Row>
+      <Row
+        css={css`
+          flex-direction: row-reverse;
+        `}
+      >
+        <Button onClick={() => {}}>Create</Button>
+      </Row>
+    </Container>
```
hlminh2000: if you update against master, there should be a `programs` page set up that can be used as a boiler plate so we don't end up with a huge file for every page. We can refactor down the road to pull the common layouts into its own thing down the road :)

client/uikit/form/Input/index.js
```diff
@@ -17,40 +18,40 @@ const Input = ({
...
-          placeholder={disabled ? '' : placeholder}
-          value={value}
-          type={type}
-          onChange={onChange}
-          size={size}
-          disabled={disabled}
-        />
-      </StyledInputWrapper>
-      {error && errorMessage ? <ErrorMsg>{errorMessage}</ErrorMsg> : null}
+      <FormControlContext.Consumer>
```
hlminh2000: We'd prefer the `useContext` API now :)

client/uikit/form/Input/index.js
```diff
@@ -17,40 +18,40 @@ const Input = ({
...
+        {contextValue => (
+          <StyledInputWrapper
+            size={size}
+            onFocus={() => setActive('focus')}
+            onBlur={() => setActive('default')}
+            error={contextValue.error}
+            disabled={contextValue.disabled}
```
hlminh2000: Ideally I'd like to have inputs remain something that's usable without the context, so maybe we can have it fallback to the props :)

client/components/NavBar.js
```diff
@@ -24,7 +24,7 @@ const NavbarLink = ({ path, active }: { path: string, active: boolean }) => {
   };
   return (
     <MenuItem
-      DomComponent={props => (
+      DomComponent={({ active, ...props }) => (
```
hlminh2000: 👍

client/uikit/form/Input/index.js
d8660091: nice

client/uikit/form/Input/index.js
```diff
@@ -17,40 +16,43 @@ const Input = ({
...
   className,
+  error,
+  disabled,
   ...props
 }) => {
   const [activeState, setActive] = useState('default');
+
+  const { disabled: calcDisabled = disabled, error: calcError = error } = useContext(
+    FormControlContext,
+  );
```
hlminh2000: do we need `useContext(FormControlContext) || {}`? what happens when there's no context?

d8660091: good catch

PR #113, opened 2019-06-14
Add new value

d8660091:


client/uikit/form/MultiSelect/index.js
```diff
@@ -110,6 +156,15 @@ function MultiSelect({ name, value, children, onChange, placeholder, inputProps
     onChange(event, child);
   };
 
+  const handleNewItemClick = event => {
+    event.target = {
+      value: _.uniq([...value, searchString]),
+      name,
```
hlminh2000: should spread the old `event.target` object into the override? if so let's do the same in `handleSelectedItemClick`

Totally not necessary if we actually don't want to expose the rest of the info, just pointing out in case it was something that was missed.

client/uikit/form/MultiSelect/stories.js
```diff
@@ -11,30 +11,54 @@ function WithState({ children }) {
...
-    </MultiSelect>
-  </WithState>
-));
-
+const MultiSelectStories = storiesOf(`${__dirname}`, module)
+  .add('Basic', () => (
+    <WithState>
+      <MultiSelect
+        value="[parent state]"
+        onChange={() => '[parent func]'}
```
hlminh2000: let's trigger the log from `@storybook/addon-actions` as well on callbacks :) it helps being able to see when and what is being dispatch form the component from the story

client/uikit/form/MultiSelect/index.js
d8660091: good point, I think it's safe 
https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Select/SelectInput.js

client/uikit/form/MultiSelect/stories.js
```diff
@@ -1,40 +1,68 @@
...
+        <Option value="Barbados">Barbados</Option>
+        <Option value="Belarus">Belarus</Option>
+      </MultiSelect>
+    </WithState>
+  ))
+  .add('Allow new value', () => (
+    <WithState>
+      <MultiSelect
+        value="[parent state]"
+        onChange={() => '[parent func]'}
```
hlminh2000: me too! :D lol

d8660091: should be resolved by https://github.com/icgc-argo/argo-platform/pull/113/commits/aaaf18f6640407d475edcd9d86f652949a1c6c49

PR #118, opened 2019-06-17
Feat/tooltip

hlminh2000:


d8660091:
Good job, but I think the second `T` in `ToolTip` should be lowercase

reference:
https://getbootstrap.com/docs/4.0/components/tooltips/
https://material-ui.com/components/tooltips/

hlminh2000:
@d8660091 great point! resolved :) thanks!

PR #122, opened 2019-06-17
updates layout of the PercentageBar

hlminh2000:
Addresses feedback in issues https://github.com/icgc-argo/argo-platform/issues/74

### DIFF
--- a/client/uikit/PercentageBar/index.js
+++ b/client/uikit/PercentageBar/index.js
@@ -4,16 +4,18 @@ import useTheme from '../utils/useTheme';
 import Typography from '../Typography';
 import Icon from '../Icon';
 import { css } from '../';
-import { withProps } from 'recompose';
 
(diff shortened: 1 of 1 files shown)


client/uikit/PercentageBar/index.js
```diff
@@ -6,7 +6,7 @@ import Icon from '../Icon';
 import { css } from '../';
 import { withProps } from 'recompose';
 
-const ValignedText = withProps(() => ({
+const VAlignedText = withProps(() => ({
```
d8660091: I heard that we deprecated recompose, didn't we?

hlminh2000: haha yes I did say that :P certainly don't want to use recompose for anything state management related, but `withProps` was a convenient function I would just write myself but didn't feel like doing so :P I'll replace with my own thing ;)

PR #128, opened 2019-06-17
Adds clear button for inputs

hlminh2000:


d8660091:
When a user hover on the close icon, it shows 'circled x sign', but the user might expect something else like clean input, doesn't he?

hlminh2000:
@d8660091 great point! addressed :)

PR #138, opened 2019-06-18
Improve multiselect component

d8660091:


hlminh2000:
instead of `single` being a prop, would it make sense to have a `limit` (int) and a `useTag` (bool)? That seems to be a more flexible API :)

d8660091:
@hlminh2000 I think `limit` is better to be handled by validation layer. Because 

- It's not tightly fit to our current design and requires additional design works. 
- `limit` is not a feature of html select element which is familiared by the millions of ordinary users and developers, they used to expect the form be validated when they click submit.
- This component is already quite complicated with 400 lines of code. Adding extra features would making it slow and hard to maintain without restructuring. 

What do you think?

PR #140, opened 2019-06-18
Sets up Apollo client with SSR and cache rehydration

hlminh2000:
partially addresses https://github.com/icgc-argo/argo-platform/issues/133

On each request, two things happen:
1) SSR server creates a server-side Apollo client to pre-fetch initial queries. Query result is used for rendering initial html response.
2) Cache from the server-side Apollo client is used to restore on a new client-side Apollo client so browser queries can pull from cached.

Next step: 
1) Create a `useApolloClient` and `useQuery` hooks for working with graphql
2) Support `.gql` files
3) Test to extract all `.gql` files for validation against remote schemas


client/components/pages/programs/ProgramsTable.js
```diff
@@ -152,3 +154,20 @@ export default (tableProps: {
   ];
   return <Table data={data} columns={columns} />;
 };
+
+/**
+ * TODO: remove this wrapper and create a `useQuery` hook to be used in ./index.js
+ */
```
hlminh2000: I didn't want to touch that file because @ciaranschutte is working on the page layout refactoring :)

joneubank:
LGTM

Issue #60, opened 2019-05-24
UI Component Library: Table Footer


kcullion:
The table footer bar can be blank, or contain pagination elements

**No pagination:**
- [ ] Just a bottom border for the table (in case the last row is white and not zebra stripe)

**With Pagination:**
![Screen Shot 2019-06-06 at 11 32 14 AM](https://user-images.githubusercontent.com/18424138/59045803-c37bae00-884e-11e9-812e-0e47a44b2edd.png)

- [x] Left side: show different number of rows per table "page"
- [x] Right side: Pagination

Zeplin:https://zpl.io/adzPEOR
See in page: https://zpl.io/aBRPxwA

**Related tickets:**
Table body: https://github.com/icgc-argo/argo-platform/issues/59
Table Action bar: https://github.com/icgc-argo/argo-platform/issues/42

hlminh2000:
@d8660091 there's already a setup for the pagination component so you can build on that, as long as we're conforming to the react-table spec ;)
Also, the TableActionBar is consumed by the page so try not to change it so much :)

kcullion:
@d8660091 I reviewed the pagination, please make the following fixes:
- [ ] The font isn't correct - it looks like a serif font - it should be WorkSans-Regular with  font-size: 12px;
- [ ] put 2px in between each item (so upon hover the circles aren't right up against eachother)
- [ ] The hover state should be on the arrows too (but not the active state)
- [ ] Let's show 5 numbers by default - and the same amount of numbers always (currently it switches from 2 numbers to 3)

PR #141, opened 2019-06-19
Fix styles issues for TablePagination

d8660091:
#60


client/uikit/Table/TablePagination/index.js
```diff
@@ -71,32 +104,37 @@ function TablePagination(props) {
         }
         & a > svg {
           position: relative;
-          top: -2px;
         }
       `}
     >
```
hlminh2000: can we just use the `TableActionBar` component above for the container so it's consistent? :) 
then we can also get rid of all the `css(theme.typography.data)` in the children

PR #146, opened 2019-06-19
use hooks in favour of Apollo's Query component

hlminh2000:
also hooks up the rest of the programs page to pull programs list with query


client/components/pages/submissionSystem/programs/index.js
```diff
@@ -47,8 +50,9 @@ export default ({
   authorizedPrograms = [],
   logOut,
   pathname,
-  programs = mockPrograms,
 }: any) => {
+  const { data, loading, errors } = useQuery(programsQuery);
+  const { programs = [] } = data;
```
d8660091: how about `const { {programs = []}, loading, errors }`?

hlminh2000: oh yeah, `const { data: { programs = [] } = {}, loading, errors } = useQuery(programsQuery);` :P

PR #152, opened 2019-06-20
✅ Static query validation test

hlminh2000:
this test validates all static graphql queries defined in `.gql` files against remote schema hosted at `process.env.API_ROOT`

Needs to be run against whatever environment that's being deployed to so kept it a separate test from all others for CI flexibility


client/package.json
```diff
@@ -40,11 +41,13 @@
     "dotenv-webpack": "^1.7.0",
     "emotion-theming": "^10.0.10",
     "express": "^4.16.4",
+    "file-regex": "^3.0.1",
     "flow-bin": "^0.98.0",
     "flow-typed": "^2.5.1",
     "graphql": "^14.2.1",
     "graphql-import": "^0.7.1",
     "graphql-tag": "^2.10.1",
+    "graphql-tools": "^4.0.4",
```
d8660091: Are file-regex and graphql-tools dependencies instead of devDependencies?

hlminh2000: good great point! thanks :)

PR #154, opened 2019-06-21
Enables Apollo tracing

hlminh2000:
tracing is good for query performance analysis

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -22,6 +22,7 @@ const init = async () => {
       egoToken: req.headers.authorization,
       dataLoaders: {},
     }),
+    introspection: true,
     tracing: NODE_ENV !== 'production',
   });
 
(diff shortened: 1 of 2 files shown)


server/app.js
```diff
@@ -22,6 +22,8 @@ const init = async () => {
       egoToken: req.headers.authorization,
       dataLoaders: {},
     }),
+    introspection: true,
+    tracing: NODE_ENV !== 'production',
```
joneubank: should we be concerned about case sensitivity here?

hlminh2000: thanks! Added a serialization step [here](https://github.com/icgc-argo/argo-platform/pull/154/files#diff-45c4a8fcc6031017d58405007de49792R2) to ensure lower case :)

server/config.js
```diff
@@ -1,4 +1,5 @@
 export const PORT = Number(process.env.PORT) || 9000;
+export const NODE_ENV = String(process.env.NODE_ENV || 'development').toLowerCase();
```
joneubank: 💯

Issue #126, opened 2019-06-17
Program Page - Read-Only Profile Tab Static UI (for Program Admin) 


rosibaj:
Program Administrators can see a read-only view of their program profile. This is not editable for them.

Zeplin: https://zpl.io/aND1kXe

![image](https://user-images.githubusercontent.com/15824252/59635985-7639ef00-9120-11e9-9448-bc7ecd82c064.png)

rosibaj:
Tested https://platform-ui.qa.argo.cancercollaboratory.org/program-management 

One feedback @d8660091 @kcullion :
![image](https://user-images.githubusercontent.com/15824252/60200458-28a23e00-9814-11e9-80d2-0bea8b617d1e.png)

kcullion:
@d8660091 

- [ ] yes the value text should be WorkSans-Regular 14px; 

- [ ] And can you try some values for the processing regions so we can get them styled too? They will come in as separate values (from checkbox component) separated by a comma.

kcullion:
nice, thanks @d8660091 !

PR #159, opened 2019-06-24
Add program-profile page

d8660091:
#126


client/components/pages/submissionSystem/program-management/index.js
```diff
@@ -0,0 +1,198 @@
...
+import Tabs, { Tab } from 'uikit/Tabs';
+import InputLabel from 'uikit/form/InputLabel';
+
+import Typography from 'uikit/Typography';
+import { Row, Col } from 'react-grid-system';
+import { ThemeContext } from '@emotion/core';
+import Icon from 'uikit/Icon';
+
+export default ({ logOut, pathname, router }) => {
+  const [activeTab, setActiveTab] = React.useState('PROFILE');
```
ciaranschutte: we could use constants for the tab states instead of strings
eg.   
`const TABS = {PROFILE : 'PROFILE', USERS: 'USERS'}`
React.useState(PROFILE)

PR #161, opened 2019-06-25
rename API_ROOT to GATEWAY_API_ROOT

ciaranschutte:
Changes name to be less generic
Updates `.env.schema` to use proper var


hlminh2000:
lgtm but will need a devops update too, so let's merge when that's updated :)

ciaranschutte:
@hlminh2000 sounds good!

PR #164, opened 2019-06-25
Modal UI component

hlminh2000:


hlminh2000:
@d8660091 yup! just added that :P integration into the app will be done in a separate PR :)

client/uikit/Modal/index.js
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { styled, css } from '..';
+import Typography from '../Typography';
+import Button from '../Button';
+import Icon from '../Icon';
+import FocusWrapper from '../FocusWrapper';
+
+const hexToRgb = hex => {
```
ciaranschutte: Might it be better to use a lib like https://polished.js.org/ or https://www.npmjs.com/package/color?
This modal is probably not the only time we need to add opacity to an item and using a lib would standardise it neatly.

```
// polished
lighten(0.2, theme.colors.primary_dark),
```

```
// color
Color(theme.colors.primary_dark).lighten(0.5);
```

hlminh2000: sure! great point :) will update

PR #170, opened 2019-06-26
Program-Service Integration through NPM

joneubank:
Also includes updates to service to match Program-Service changes, and introduces invite user method


joneubank:
https://app.zenhub.com/workspaces/argo-5c8c8637777e5f767e486217/issues/icgc-argo/argo-platform/97

server/schemas/Program/index.js
```diff
@@ -167,9 +138,16 @@ const resolvers = {
...
+      const program = get(args, 'program', {});
       const response = await programService.createProgram(program, egoToken);
-      return get(response, 'id.value');
+      return get(args, 'program.shortName');
+    },
+    inviteUser: async (obj, args, context, info) => {
+      const { egoToken } = context;
+      const invite = get(args, 'invite', {});
+      const response = await programService.inviteUser(invite, egoToken);
+      console.log(response);
```
hlminh2000: 🔥

server/schemas/Program/index.js
```diff
@@ -93,69 +84,49 @@ const typeDefs = gql`
...
+    Returns the shortName of the program if successfully created
+    """
+    createProgram(program: ProgramInput): String
+
+    """
+    Invite a user to join a program
+    Returns the email of the user if the invite is successfully sent
     """
-    createProgram(program: ProgramInput): String!
+    inviteUser(invite: InviteUserInput): String
```
hlminh2000: maybe type this as `InviteUserInput!`?

server/schemas/Program/index.js
```diff
@@ -93,69 +84,49 @@ const typeDefs = gql`
   }
   type Mutation {
     """
-    create new program
+    Create new program
+    Returns the shortName of the program if successfully created
+    """
+    createProgram(program: ProgramInput): String
```
hlminh2000: `ProgramInput!`?

hlminh2000: Also let's return a `Program`? Else UI will have to take that shortName to fetch user again separately anyway.
If we want to optimize the resolver later, we can also do some lookahead to only make the upstream read if fields other than the `shortName` are requested.

server/schemas/Program/index.js
joneubank: yo that shouldnt be there

hlminh2000: :P

server/schemas/Program/index.js
joneubank: program created will have no info that wasnt in the input... is this valuable to return?

hlminh2000: Ah that makes sense.
Though I think it's easier to have it now than potentially have to introduce breaking change in future, in case we add computed properties and/or relationship joins in later.

joneubank: Would require an extra fetch to Program-Service for this info. Nothing is returned after successful create.

Issue #51, opened 2019-05-22
UI Component Library: Footer


rosibaj:
- [ ] Desktop footer: content and links TBD
- [x] OICR logo should open oicr.on.ca in a new tab

Zeplin: https://zpl.io/adzPEOR
See in page: https://zpl.io/aBRPxwA

![Screen Shot 2019-06-25 at 4 00 56 PM](https://user-images.githubusercontent.com/18424138/60129187-6fcff680-9762-11e9-9b94-b624e3f520cf.png)

Related issue:
Mobile Footer: https://github.com/icgc-argo/argo-platform/issues/61

kcullion:
@d8660091 
please make the following changes:
- [ ] put "All rights reserved." on top line with copyright. don't capitalize "rights"
- [ ] change overall footer hight to 64px;
- [ ] horizontally center the middle grouping of links
- [ ] add target="_blank" to the oicr logo so it opens in a new tab

![argo-footer-fixes](https://user-images.githubusercontent.com/18424138/60274871-e1c15080-98c6-11e9-81bc-a798dbb036e7.jpg)

kcullion:
@d8660091 just a couple more things:

- [ ] Use the svg file for the OICR logo (in any case for vector images, please implement svgs)
- [ ] make the OICR logo slightly smaller (height:46px) and vertically centered within the footer. 
- [ ] include top and bottom padding (8px) so the height isn't dependent for the footer so there is breathing space for the logo
![Screen Shot 2019-06-27 at 1 42 44 PM](https://user-images.githubusercontent.com/18424138/60288564-771e0e00-98e2-11e9-8e11-2a45b59c6d08.png)
- [ ] also in the screenshot above - I think something with the footer is making the page have a horizontal scrollbar. Please fix

d8660091:
@kcullion current logo is img with svg, isn't it

kcullion:
@d8660091 I'm sorry, I saw it wrong!

So have a look at the accessiblity tab in storybook - there are some violations (maybe to do with the empty links though)
The OICR logo should have an alt tag though.  Use "Ontario Institute for Cancer Research" please

kcullion:
@d8660091 can you just do this one checkbox so the accessibility violation is fixed :)

- [ ] The OICR logo should have the alt tag: "Ontario Institute for Cancer Research"

rosibaj:
Tested - alt tag is fixed thanks @d8660091  !!

PR #174, opened 2019-06-26
Add footer 

d8660091:
#51


hlminh2000:
wanna bing it into the `client/components/pages/submissionSystem/layout.js` too? could be done in a separate PR, up to you :)

d8660091:
> wanna bing it into the `client/components/pages/submissionSystem/layout.js` too? could be done in a separate PR, up to you :)

I'll try to do that in this pull-request

PR #195, opened 2019-07-02
adds dataloader to programs query

hlminh2000:
```graphql
{
  request0: programs {
    shortName
    description
  }
  request1: programs {
    shortName
    description
  }
  request2: programs {
    shortName
    description
  }
  ...
}
```
For the above query, data loader will allow gateway to only hit upstream service once.


rosibaj:
Closing PR as it will be handled in a different implementation (referenced above)

PR #196, opened 2019-07-02
adds query complexity limit to gateway

hlminh2000:


server/app.js
```diff
@@ -24,6 +26,13 @@ const init = async () => {
     }),
     introspection: true,
     tracing: NODE_ENV !== 'production',
+    validationRules: [
+      costAnalysis({
+        maximumCost: GQL_MAX_COST,
+        // logs out complexity so we can later on come back and decide on appropriate limit
+        onComplete: cost => console.log(`QUERY_COST: ${cost}`),
```
joneubank: if we are logging complexity, can we also log the request that generated it?

hlminh2000: I was really hoping that was possible, but unfortunately we don't get access to the request here T_T 
I hope they support it in newer versions of apollo

joneubank: well ok then :P . seperately, we need to log requests and add requests id's / time stamps through a logging framework. another issue.

PR #203, opened 2019-07-02
Feat/#115 create program layout

ciaranschutte:


client/pages/create-program.js
```diff
@@ -231,12 +239,27 @@ export default createPage({
...
+            justify-content: space-between;
+            flex: 1;
+            display: flex;
+          `}
+        >
+          <Button
+            variant="secondary"
+            css={css`
+              border: none;
+            `}
```
hlminh2000: if we get this (https://github.com/icgc-argo/argo-platform/pull/198) merged, we can now do:
```
<Button variant="text" ... />
```
so no need to set `border: none;`

ciaranschutte:
@hlminh2000 moved `create-program` into page components, left page file as lightweight router

PR #213, opened 2019-07-03
egoJwt sending properly

hlminh2000:


server/app.js
```diff
@@ -21,7 +21,7 @@ const init = async () => {
     }),
     context: ({ req }) => ({
       isUserRequest: true,
-      egoToken: req.headers.authorization,
+      egoToken: req.headers.authorization.split('Bearer ').join(''),
```
hlminh2000: actually should be `(req.headers.authorization || "").split('Bearer ').join('')`

PR #214, opened 2019-07-04
Feat/#53 add users static

ciaranschutte:
Adds the static UI for add user modal


client/components/pages/submissionSystem/modals/addUser/formSchema.js
```diff
@@ -0,0 +1,19 @@
+import * as yup from 'yup';
+
+const testGsuite = async email => {
```
joneubank: Its a good placeholder... have you figured out how to actually confirm this yet?

client/components/pages/submissionSystem/modals/addUser/formSchema.js
```diff
@@ -0,0 +1,19 @@
...
+  return true;
+};
+
+const AddUserSchema = yup.object().shape({
+  firstName: yup.string().required('Please fill in first name'),
+  lastName: yup.string().required('Please fill in last name'),
+  email: yup
+    .string()
+    .required('Please fill in email')
+    .test('is-gsuite', 'email is not gsuite', testGsuite),
```
joneubank: A note about ordering the validations: 
We want to validate the form with the `abortEarly:false` option so that we find all the failures, not just the first. When this is done, the message that will be returned is for the each field will be from the last validation in the list that failed. In this case, if email is blank, it will fail the required validation, and the is-gsuite validation, and the messge returned would be "email is not gsuite". This would look weird since the email field is empty.

TL;DR - Put required as the last validation in the list.

client/components/pages/submissionSystem/modals/addUser/index.js
```diff
@@ -0,0 +1,204 @@
...
+import Button from 'uikit/Button';
+import Icon from 'uikit/Icon';
+import Typography from 'uikit/Typography';
+import css from '@emotion/css';
+import UserSection from './userSection';
+import AddUserSchema from './formSchema';
+import * as yup from 'yup';
+import { get } from 'lodash';
+
+const useFormHook = ({ initialFields, schema: formSchema }) => {
```
joneubank: longterm we can abstract this to use in both this form and create-program. this has all the boilerplate state management i wrote inline in the form component.

client/components/pages/submissionSystem/modals/addUser/formSchema.js
ciaranschutte: https://app.zenhub.com/workspaces/argo-5c8c8637777e5f767e486217/issues/icgc-argo/argo-platform/50

apparently you can check against the DNS record... the specifics of this I do not know yet

client/components/pages/submissionSystem/modals/addUser/index.js
ciaranschutte: Sounds good. This is still a little too rigid. for instance it assumes each "section" is of the type "formSchema"

client/components/pages/submissionSystem/modals/addUser/formSchema.js
ciaranschutte: great! nice catch

joneubank: Also if you provide a `.label("Email")` you can skip the error message for required... the default text is good enough

client/uikit/form/Input/index.js
```diff
@@ -43,7 +44,10 @@ const Input = ({
       <StyledInputWrapper
         size={size}
         onFocus={() => setActive('focus')}
-        onBlur={() => setActive('default')}
+        onBlur={() => {
+          setActive('default');
+          onBlur();
```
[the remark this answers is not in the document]
ciaranschutte: covered in Jon's merged PR :)

client/uikit/Icon/icons/index.js
```diff
@@ -18,6 +18,7 @@ import warning from './collection/warning';
 import mail from './collection/mail';
 import trash from './collection/trash';
 import google from './collection/google';
+import plus_circle from './collection/plus_circle';
```
d8660091: Do we have any name convention for js files?

client/components/pages/submissionSystem/modals/addUser/index.js
```diff
@@ -0,0 +1,227 @@
...
+  const [form, setForm] = useState({ errors: [initialFields], data: [initialFields] });
+  const [touched, setTouched] = useState(false);
+  const { errors, data } = form;
+
+  const hasErrors = errors
+    .map(section => Object.values(section))
+    .flat()
+    .reduce((acc, val) => acc || !!val, false);
```
d8660091: Array.prototype.some() is an alternative

client/components/pages/submissionSystem/modals/addUser/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+  // set form data
+  const setData = ({ key, val, index }) => {
+    if (!touched) setTouched(true);
+
+    setForm({
+      ...form,
+      data: data.map((field, i) => (i === index ? { ...field, [key]: val } : field)),
```
d8660091: An alternative: _.set(data, `[${index}].${key}`, val)

client/uikit/Icon/icons/index.js
ciaranschutte: I just continued what was there already: use an underscore to split words

client/components/pages/submissionSystem/modals/addUser/userSection.js
```diff
@@ -0,0 +1,126 @@
...
+
+  return (
+    <Section>
+      <div
+        css={css`
+          flex: 1;
+        `}
+      >
+        <Row nogutter>
+          <Col sm={6} style={{ paddingRight: '5px' }}>
```
hlminh2000: these paddings are causes a slight shift in the layout on smaller screen:
Screen Shot 2019-07-11 at 3 57 44 PM

ciaranschutte: oof, good catch thank you

client/components/pages/submissionSystem/modals/addUser/index.js
ciaranschutte: I think _.set returns same object and we need it immutable for setting new state

d8660091: how about _.set({...data})?

PR #216, opened 2019-07-04
adds link with chevron config

hlminh2000:


client/uikit/Link/index.js
```diff
@@ -19,16 +25,33 @@ const StyledLink = styled('span')`
...
 
-const Link = ({ href, Component = BuiltInDomLink, ...rest }) => (
-  <StyledLink>
-    <Component href={href} {...rest} />
+const Link = ({
+  href,
+  Component = BuiltInDomLink,
+  variant = LINK_VARIANTS.INLINE,
+  uppercase = variant === LINK_VARIANTS.BLOCK,
+  withChevron = variant === LINK_VARIANTS.BLOCK,
```
ciaranschutte: this is cool. I didn't know you could destructure with an already destructed val in same expression.

Is this confusing that our default arg isn't static and in saying that... not really a default arg? @hlminh2000

hlminh2000: We've actually been doing this quite a bit. It's a native javascript thing so I think it's quite idiomatic :)

We could have these defined as local variables computed based on props, but that would would be more confusing imo because that behaviour is not surfaced to the API (not picked up by the storybook proptypes table)

ciaranschutte: sounds good!

PR #217, opened 2019-07-04
add components storybook section

ciaranschutte:


hlminh2000:
could we add a sample story? :)

client/stories/comp.stories.js
```diff
@@ -0,0 +1,14 @@
...
+
+import { ThemeProvider } from '../uikit';
+
+const req = require.context('../components', true, /.stories\.js$/);
+
+function loadStories() {
+  req.keys().forEach(filename => req(filename));
+}
+
+addDecorator(story => <ThemeProvider>{story()}</ThemeProvider>);
```
hlminh2000: This will make react hooks work in storybook :D
```
addDecorator(Story => <ThemeProvider><Story/></ThemeProvider>);
```
Out of scope but let's change this in the `uikit` storybook too. Will simplify writing stories a lot :)

PR #222, opened 2019-07-05
Feat/ui access control

hlminh2000:
Not complete, but since it touches a lot of things, I wanted to get this merged incrementally to avoid conflict


client/pages/_app.js
```diff
@@ -112,7 +113,7 @@ const Root = (props: {
                   <pre>{error.stack || error.message}</pre>
                 )
               ) : unauthorized ? (
-                'You are not authorized'
+                <Error401Page />
```
d8660091: This error can be handled by _error.js
```
  const err = new Error('An Expected error occurred');
  err.statusCode = 401;
  throw err;
```

hlminh2000: great point! thanks!

hlminh2000: @d8660091 I had to tweak the Error page a little to get the error key picked up properly cause but so if you could take a look there that'd be awesome ;)

PR #224, opened 2019-07-08
Add error pages

d8660091:


client/components/pages/submissionSystem/error.js
```diff
@@ -0,0 +1,54 @@
...
+        `}
+        links={[
+          { displayName: 'Contact', href: '#' },
+          { displayName: 'Documentation', href: '#' },
+          { displayName: 'Privacy Policy', href: '#' },
+        ]}
+      />
+    </div>
+  );
+}
```
hlminh2000: let's move this outside of the `submissionSystem` folder, I don't think it's specifically for the submission system.

PR #228, opened 2019-07-08
Add program profile data

d8660091:


client/components/pages/submissionSystem/program-management/index.js
```diff
@@ -15,8 +18,32 @@ import Button from 'uikit/Button';
...
+        website
+        institutions
+        countries
+        regions
+        membershipType
+        cancerTypes
+        primarySites
+      }
+    }
+  `;
```
hlminh2000: Let's put this query in `./queries.gql` and provide the `shortName` through `variables`.
That allows the query to be picked up by our contract test :)

client/pages/program-management/[shortName].js
```diff
@@ -1,15 +1,10 @@
...
 import ProgramManagement from 'components/pages/submissionSystem/program-management';
 
-// $FlowFixMe .gql file not supported
-import { programsListQuery } from 'components/pages/submissionSystem/programs/queries.gql';
-
 export default createPage({
   isPublic: true,
   isAccessible: async ({ egoJwt, ctx }) => !isRdpcMember(egoJwt),
-  getGqlQueriesToPrefetch: async () => [{ query: programsListQuery }],
 })(ProgramManagement);
```
Comment (hlminh2000) [inline client/pages/program-management/[shortName].js]: Let's make the path `/program/manage/[shortName]`. 
There'll be other pages with similar scope so, easier to do stuff programatically later on :)

PR #229, opened 2019-07-08
Create Program Form

joneubank:
Includes UI styling, validation, submission.


joneubank:
Needs a bit of refactor, opening so you can take a look at form stuff, like error handling, validation, submission.

joneubank:
Ready for final review then merge:
* Refactored validation schema to its own file
* Split constants into pages and general files
* Added constants for create program form fields, form and validations refer to these lists now.
* Added field validations on blur that update the error messages

joneubank:
Also addressed comments left by Minh previously ;)

ciaranschutte:
Awesome work. Pretty close to my form...except more awesome. 

I ended up throwing all my form state and validation in a hook but we can decide about how we want to standardise forms as we work.

client/components/pages/submissionSystem/programs/index.js
```diff
@@ -63,7 +65,9 @@ export default ({
           >
             All Programs
           </Typography>
-          <Button onClick={console.log}>Create a program</Button>
+          <Link href="/create-program">
```
hlminh2000: Let's pull this path from a global constant :)

client/components/pages/submissionSystem/create-program/CreateProgramForm.js
```diff
@@ -0,0 +1,553 @@
...
+  const validateField = (path, value) => {
+    yup
+      .reach(createProgramSchema, path)
+      .validate(value, { abortEarly: false })
+      .then(success => {
+        updateValidationErrorsForField(path, '');
+      })
+      .catch(err => {
+        console.log(err);
+        const message = err.inner ? err.inner[err.inner.length - 1].message : err.message;
```
ciaranschutte: [ ] is truthy so it's gonna try and get the inner message of an empty array potentially @joneubank

joneubank: yeah true. worse... it has a value even when there is only one error.
I can make it check for existence and length >1 instead.

client/pages/create-program.js
```diff
@@ -1,20 +1,13 @@
 import { createPage } from 'global/utils/pages';
-import CreateProgram from 'components/pages/submissionSystem/create-program';
+
+import CreateProgramPage from '../components/pages/submissionSystem/create-program';
```
hlminh2000: did the absolute path not work? just curious

client/uikit/Typography/index.js
```diff
@@ -31,7 +32,7 @@ const createDomComponent = memoize((domComponentName, components, variant) =>
 
 const createStyledDomComponent = memoize(
   Component => styled(Component)`
-    font-weight: ${({ bold }) => (bold ? `bold` : `normal`)};
+    ${({ bold }) => (bold ? `font-weight:bold` : ``)};
```
hlminh2000: This would cause some variants to not have a non-bold version. We're specifying `bold` in the `SectionTitle` component anyway, so I think we can do without this change

client/uikit/form/Select/index.js
```diff
@@ -64,6 +65,7 @@ const Select = ({
         onBlur={() => {
           setActive('default');
           setExpanded(false);
+          onBlur();
```
hlminh2000: do we care to pass the `event` object up too? 
also a quick update to storybook to show the `onBlur` would be great :)

client/pages/create-program.js
joneubank: absolute paths work. don't know why i did this. nice catch.

client/components/pages/submissionSystem/programs/index.js
joneubank: oh yeah the constants I specifically organized...

client/uikit/Typography/index.js
joneubank: I can revert this but we still need to work out our typography styling for the future. We are specifying font weight there but this overwrites it in all cases.

client/uikit/form/Select/index.js
joneubank: Yeup, both are good requests.

client/uikit/Typography/index.js
hlminh2000: I think the most consistent way to model the Kim's idea is to remove the `fontWeight` from the theme and just let this handle it. We can decide with her on a set of commonly used configurations and make components when she's back :)

PR #231, opened 2019-07-09
Fix display issues of program management page

d8660091:


client/components/pages/submissionSystem/program-management/index.js
```diff
@@ -134,21 +136,29 @@ function Profile({ program = {} }) {
   const theme = React.useContext(ThemeContext);
   const Left = props => (
     <Col
-      md={3}
+      md={2}
```
hlminh2000: This is out of scope but it's probably easier to just do it here than wasting more ticket cycle... so if you feel like it, maybe add a `sm={12}` to these columns so they flow down earlier to avoid this overlap on small screen :)

Screen Shot 2019-07-10 at 9 09 27 AM

Only if it doesn't break anything else though ;)

PR #232, opened 2019-07-09
Clean up header

d8660091:


client/components/pages/submissionSystem/NavBar.js
```diff
@@ -1,19 +1,27 @@
...
+import { css } from 'uikit';
 import AppBar, { Logo, MenuGroup, MenuItem, Section, UserBadge } from 'uikit/AppBar';
 import Button from 'uikit/Button';
-import { css } from 'uikit';
+
+const submissionPaths = [
+  PROGRAMS_LIST_PATH,
+  PROGRAM_ENTITY_PATH,
+  DCC_OVERVIEW_PATH,
+  PROGRAM_MANAGE_PATH + '/[shortName]',
```
hlminh2000: Does this work with `urlJoin`? want to avoid any potential double slashes (`//`) :)

client/components/pages/submissionSystem/NavBar.js
```diff
@@ -77,7 +82,7 @@ export default (props: { path: string, logOut: void => void, children?: React.No
       <Section />
       <Section>
         <MenuGroup>
-          <MenuItem>Submission System</MenuItem>
+          <MenuItem active={_.includes(submissionPaths, props.path)}>Submission</MenuItem>
```
hlminh2000: `"/program/panage/PACA-AU"` wouldn't match the value of `PROGRAM_MANAGE_PATH`. I think we can bring `submissionPaths` into this component and construct the case for `/program/manage/` dynamically with `urlJoin` and `PROGRAM_MANAGE_PATH = '/program/manage/`

d8660091: Yes, it would match. `props.path` of "/program/panage/PACA-AU" would be "/program/panage/[shortName]"

hlminh2000: ah! that's how Next passes it eh :) that's awesome, thanks @d8660091 !

PR #241, opened 2019-07-10
Add error handling

d8660091:


client/pages/_error.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Page404 from './404';
+import Page500 from './404';
```
hlminh2000: we won't actually have `/404` and `/500` pages so let's import these from `/components` :)

PR #248, opened 2019-07-11
CreateProgram interface add admin fields

joneubank:
UI and Server updates for Create Program changes in ProgramServiceProto 0.0.2 - adds admin first and last names into admins list.


joneubank:
Also took the chance to add better front end validations to shortName

PR #250, opened 2019-07-11
authorization for the program entity pages

hlminh2000:
- moves all the submission related pages into the `/submission` path.
- implements program admin authorization for the program entity pages
- more path formalization


client/global/constants/pages.js
```diff
@@ -1,10 +1,12 @@
...
-export const PROGRAM_MANAGE_PATH = '/program/manage/[shortName]';
-export const PROGRAM_DASHBOARD_PATH = '/program/dashboard/[shortName]';
-export const PROGRAM_ID_REGISTRATION_PATH = '/program/id-registration/[shortName]';
-export const PROGRAM_CLINICAL_SUBMISSION_PATH = '/program/clinical-submission/[shortName]';
+
+// submission paths
+export const SUBMISSION_PATH = `/submission`;
+export const PROGRAMS_LIST_PATH = `${SUBMISSION_PATH}/program`;
+export const PROGRAM_SHORT_NAME_PATH = `[shortName]`;
+export const CREATE_PROGRAM_PAGE_PATH = `${SUBMISSION_PATH}/program/create-program`;
```
joneubank: The replace stuff works well, just having a thought...

What if these constants were instead functions that required a short name? Ex: ```export const PROGRAM_MANAGE_PATH = shortName => `${SUBMISSION_PATH}/program/${shortName}/manage`;```

Or we keep the constants and also export these convenience methods from here... Is that better than requiring someone to know that they need to replace PROGRAM_SHORT_NAME_PATH every time they use the constant?

hlminh2000: Thanks!
I do like the function idea, just wasn't sure how to organize it. `next` actually passes us `"/submission/program/[shortName]/manage"` so figured there's probably value in having a way to access the version with the `"[shortName]"` part. 
Wasn't something I have an answer for and didn't want it to block the breaking change for now :P

PR #253, opened 2019-07-12
Update logged in page

d8660091:


client/pages/logged-in.js
```diff
@@ -7,6 +7,9 @@ import { LOCAL_STORAGE_REDIRECT_KEY } from 'global/constants';
 import { LOGIN_PAGE_PATH } from 'global/constants/pages';
 import { getRedirectPathForUser } from 'global/utils/pages';
 import { createPage } from 'global/utils/pages';
+import { css } from 'uikit';
+import colors from 'uikit/theme/defaultTheme/colors';
```
hlminh2000: let's use the `useTheme` hook from uikit instead of importing themes directly ;)
```javascript
import useTheme from 'uikit/utils/useTheme';

// then
const theme = useTheme()
```

PR #258, opened 2019-07-12
Feat/#50 add users data hookup

ciaranschutte:


client/.storybook/webpack.config.js
```diff
@@ -1,19 +1,19 @@
...
   };
 
   config.resolve.modules.push(path.resolve(__dirname, '../'));
 
+  config.module.rules.push({
+    test: /\.(graphql|gql)$/,
+    exclude: /node_modules/,
+    loader: 'graphql-tag/loader',
+  });
+
```
hlminh2000: In an ideal world, this is awesome :P but I think we want to hold off having any data tied into stories for now. 
Let's limit any data hook up to be done at application level and fed into UI components :)

ciaranschutte: @hlminh2000 that makes sense. I actually just added this so Storybook webpack wouldn't complain about our .gql files. But I think I write a webpack exclude rule instead

ciaranschutte: @hlminh2000 actually looks like that won't work
`addUser` imports the `mutations.gql` so still needs to load

client/components/pages/submission-system/modals/addUser/index.js
```diff
@@ -152,13 +155,29 @@ const AddUserModal = ({}) => {
...
+      userLastName: data.lastName,
+      userEmail: data.email,
+      userRole: data.role,
+    };
+  };
+
+  const sendAddUser = validData =>
+    useMutation(INVITE_USER_MUTATION, {
+      variables: { user: createUserInput(validData) },
+    });
```
hlminh2000: let's move this outside of this hook and define it in the page, fed to the component as `onSubmit` or something :) that way we don't expose anything to the story

ciaranschutte:
@hlminh2000 @joneubank had to update react-apollo-hooks, so believe I broke create-program form. will fix it in this: https://app.zenhub.com/workspaces/argo-5c8c8637777e5f767e486217/issues/icgc-argo/argo-platform/307

client/components/pages/submission-system/modals/addUser/index.js
```diff
@@ -40,13 +39,14 @@ const AddUserModal = ({ dismissModal }) => {
   const submitForm = async () => {
     try {
       const validData = await validateForm();
-      console.log(validData);
-      // Send data
+      const result = sendAddUser(validData);
```
hlminh2000: I think we can just do `onSubmit(validData)` here and remove the `sendAddUser` function eh?

client/.storybook/webpack.config.js
hlminh2000: is this still needed?

ciaranschutte:
@hlminh2000 removed graphql tag and extra func

PR #261, opened 2019-07-12
Feat/ego token utils

hlminh2000:
still keeping this file for now for flow type before we move to typescript


client/global/utils/egoJwt/index.js
```diff
@@ -47,107 +32,30 @@ type EgoJwtData = {
 
 type PermissionScopeObj = {
   policy: string,
-  permission: $Values<typeof PERMISSIONS>,
+  permission: 'READ' | 'WRITE' | 'ADMIN' | 'DENY',
```
joneubank: From enum?

PR #262, opened 2019-07-15
🚧WIP - Typescript!

hlminh2000:
I'll just keep rebasing this against master as we go, will make formalize when things fully work


client/uikit/form/Input/index.js
```diff
@@ -14,18 +14,18 @@ export const INPUT_PRESETS = {
...
 const Input = ({
   preset = INPUT_PRESETS.DEFAULT,
-  value,
-  onChange,
-  onBlur = () => {},
-  type,
+  value = undefined,
+  onChange = e => {},
+  onBlur = e => {},
+  type = undefined,
```
hlminh2000: providing default tells TS that it's optional without having to type it

PR #265, opened 2019-07-15
Add contact page

d8660091:


client/components/pages/contact.js
```diff
@@ -0,0 +1,357 @@
...
+            <Typography
+              css={css`
+                margin: 0;
+                margin-bottom: 42px;
+              `}
+            >
+              Check out these common topics before submitting a question below.
+            </Typography>
+          </div>
+          <div>
```
hlminh2000: why don't we use `<Col />` to account for screen sizes here?

d8660091: Col could be used in this case, but I think less elegant because the dealing with margins in this case. And display on small screen is not a big concern, is it?

hlminh2000: `<Col/>` with a horizontal padding, or margin on the children doesn't sound too bad.

PR looks good otherwise so can address after, but these public pages are more likely to be seen on mobile so we might as well for these low-hanging fruits

d8660091: sounds good

PR #266, opened 2019-07-15
Access control for submission pages

hlminh2000:


client/components/pages/submission-system/NavBar.js
```diff
@@ -57,6 +57,9 @@ export default (props: { path?: string, logOut: void => void, children?: React.N
     }
   })();
 
+  const canAccessSubmission =
+    !!egoJwt && (canReadProgram({ egoJwt, programId: '' }) || isRdpcMember(egoJwt));
```
joneubank: Is this correct? how do you check permission without the programId?

joneubank: If this is the syntax for ability to read ALL programs... we should maybe make an explicit method for that check

client/components/pages/submission-system/programs/index.js
```diff
@@ -85,9 +86,11 @@ export default ({
           >
             All Programs
           </Typography>
-          <Link href={CREATE_PROGRAM_PAGE_PATH}>
-            <Button>Create a program</Button>
-          </Link>
+          {isDccMember(egoJwt) && (
+            <Link href={CREATE_PROGRAM_PAGE_PATH}>
```
joneubank: 👍

client/package.json
```diff
@@ -27,7 +27,7 @@
     "@emotion/babel-preset-css-prop": "^10.0.9",
     "@emotion/core": "^10.0.10",
     "@emotion/styled": "^10.0.11",
-    "@icgc-argo/ego-token-utils": "^3.0.0",
+    "@icgc-argo/ego-token-utils": "^3.1.0",
```
joneubank: We're at version 3 already?!

client/components/pages/submission-system/NavBar.js
hlminh2000: lol yes :))). Alternatiely, maybe default `programId` to `""`, then run `canReadProgram({ egoJwt })`? it's basically ability to read any program at all :P

client/package.json
hlminh2000: 1 and 2 were my practice releases that failed lol

client/components/pages/submission-system/NavBar.js
joneubank: Yeah lets make it that for the future, and document the special case. What I don't want is a partial string match against programID being done inside the ego token utils. Should be exact match against program or empty string check for any program.

joneubank: on second though, after our outside discussion... lets have a separate method for this check, not a special case of canReadProgram()

client/pages/submission/program/index.js
```diff
@@ -10,6 +10,7 @@ import { programsListQuery } from 'components/pages/submission-system/programs/q
 
 export default createPage({
   isPublic: false,
-  isAccessible: async ({ egoJwt, ctx }) => !isRdpcMember(egoJwt),
+  isAccessible: async ({ egoJwt, ctx }) =>
+    !isRdpcMember(egoJwt) && canReadProgram({ egoJwt, programId: '' }),
```
joneubank: Can you clarify the rule here for me? RDPC users can't access program list? What about DCCMember (not included in this access rule)

hlminh2000: `canReadProgram` already handles `dcc` users

hlminh2000: and yes, `RDPC` can't access program list, as per https://wiki.oicr.on.ca/display/icgcargotech/Sitemap#Sitemap-PermissionsMatrixforSiteMap

joneubank: thanks. the ego token utils changes we discussed will make this clearer for future generations. 👩‍🚀

PR #273, opened 2019-07-16
UpdateProgram in Gateway

joneubank:
There is a hack to deal with a peculiarity of the UpdateProgram service: Before calling update, we call GetProgram to find the current values. The call to UpdateProgram provides all those values for fields that are not desired to be overwritten.

Otherwise, normal gateway service patternz...

### DIFF
--- a/server/schemas/Program/index.js
+++ b/server/schemas/Program/index.js
@@ -76,6 +76,8 @@ const typeDefs = gql`
   }
 
   input UpdateProgramInput {
+    # This intentionally does not provide access to submittedDonors or genomicDonors
+    # Those are maintained by an internal service and should not be updated by any client through the gateway
     name: String
     description: String
     commitmentDonors: Int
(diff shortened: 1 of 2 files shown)


joneubank:
I opened this bug on the program service https://github.com/icgc-argo/program-service/issues/145

It is not blocking this PR or issue as I have a workaround in place, but we should address that bug so we can simplify the implementation here.

server/schemas/Program/index.js
```diff
@@ -75,6 +75,21 @@ const typeDefs = gql`
     admins: [ProgramUserInput!]!
   }
 
+  input UpdateProgramInput {
+    # This intentionally does not provide access to submittedDonors or genomicDonors
+    # Those are maintained by an internal service and should not be updated by any client through the gateway
```
hlminh2000: 👍

PR #282, opened 2019-07-16
Feat/app modal

ciaranschutte:
adds a top level AppModal portal

### DIFF
--- a/client/components/pages/submission-system/layout.js
+++ b/client/components/pages/submission-system/layout.js
@@ -16,6 +16,7 @@ import Head from '../head';
 import NavBar from './NavBar';
 import SideMenu from './SideMenu';
 import Footer from 'uikit/Footer';
+import Modal from 'uikit/Modal';
 
 /**
  * TODO: `pathname` and `logOut` should just be available through context
@@ -24,7 +25,7 @@ const modalPortalRef = React.createRef();
 
 export const AppModal = ({ children }: { children: React.Node }) => {
   const ref = modalPortalRef.current;
-  return ref ? ReactDOM.createPortal(<div>{children}</div>, ref) : null;
+  return ref ? ReactDOM.createPortal(<Modal.Overlay>{children}</Modal.Overlay>, ref) : null;
 };
 
 const SubmissionLayout = ({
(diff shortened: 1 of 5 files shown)


hlminh2000:
does it make sense to bundle `Modal.Overlay` with `AppModal`?

client/components/pages/submission-system/layout.js
```diff
@@ -36,21 +45,29 @@ const SubmissionLayout = ({
...
+          {contentHeader && <ContentHeader>{contentHeader}</ContentHeader>}
+          <ContentBody>{children}</ContentBody>
+        </PageContent>
+      </PageBody>
+      <PageFooter>
+        <Footer />
+      </PageFooter>
+      <div
+        css={css`
+          position: absolute;
```
hlminh2000: would `fix` position be better?

client/components/pages/submission-system/layout.js
```diff
@@ -18,6 +20,13 @@ import Footer from 'uikit/Footer';
 /**
  * TODO: `pathname` and `logOut` should just be available through context
  */
+const modalPortalRef = React.createRef();
+
+export const AppModal = ({ children }: { children: React.Node }) => {
+  const ref = modalPortalRef.current;
+  return ref ? ReactDOM.createPortal(<div>{children}</div>, ref) : null;
```
hlminh2000: instead of `<div>` I guess we can do `<Modal.Overlay>` eh?

client/components/pages/submission-system/layout.js
```diff
@@ -18,6 +20,13 @@ import Footer from 'uikit/Footer';
 /**
  * TODO: `pathname` and `logOut` should just be available through context
  */
+const modalPortalRef = React.createRef();
+
+export const AppModal = ({ children }: { children: React.Node }) => {
```
hlminh2000: maybe let's name this `ModalPortal`? :)

client/components/pages/submission-system/layout.js
ciaranschutte: fixed will leave the sidebar and footer over it

PR #310, opened 2019-07-18
adds link to all programs list page from navbar

hlminh2000:


client/components/pages/submission-system/SideMenu.js
```diff
@@ -163,7 +171,13 @@ export default () => {
...
 
   return (
-    <Submenu>
+    <Submenu
+      css={css`
+        & a {
+          text-decoration: none;
+        }
+      `}
+    >
```
ciaranschutte: Should this just be in the Submenu uikit component? Is it an exception case or the standard?

hlminh2000: hmm I think you're right, it should be the standard. Having an underline would be the exception for this :)

PR #311, opened 2019-07-18
adds user role in the nav bar

hlminh2000:


client/components/pages/submission-system/NavBar.js
```diff
@@ -58,6 +64,15 @@ export default (props: { path?: string, logOut: void => void, children?: React.N
   })();
 
   const canAccessSubmission = !!egoJwt && (canReadSomeProgram(egoJwt) || isRdpcMember(egoJwt));
+  const userRole = !egoJwt
```
joneubank: Nested ternaries can be a pain to maintain...

Would you consider making a `getUserRole()` method in this file and put a switch or if/else ladder? Would also make it easy to separate from this file in the future if we need this logic in other places in the future...

hlminh2000: I don't quite know how to write a switch anymore :P have converted to if chain :) Thanks!

PR #314, opened 2019-07-18
implements default login redirect based on user token

hlminh2000:


client/components/pages/submission-system/NavBar.js
```diff
@@ -80,8 +82,20 @@ export default (props: { path?: string, logOut: void => void, children?: React.N
...
         <MenuGroup>
-          {canAccessSubmission && (
-            <MenuItem active={path.search(SUBMISSION_PATH) === 0}>Submission</MenuItem>
+          {egoJwt && canAccessSubmission && (
+            <MenuItem
+              active={path.search(SUBMISSION_PATH) === 0}
+              DomComponent={props => (
+                <Link
+                  href={getDefaultRedirectPathForUser(egoJwt)}
+                  as={getDefaultRedirectPathForUser(egoJwt)}
```
d8660091: Is `as` redundant?

hlminh2000: It's for the `prefetch` to work, but you're right, the `as` should be slightly different. I forgot to update this part, will do! :P

PR #315, opened 2019-07-18
Address feedbacks of program editing page

d8660091:


joneubank:
@d8660091 Can you add a description or comment with an list of all the changes included in thiss PR?

There are several changes included here and I'm struggling to identify all of the new functionality vs what was asked for... This is partly because the linked ticket isn't very clear about what has been completed before and what is new in this request.

.github/pull_request_template.md
```diff
@@ -11,9 +11,10 @@ Please add a brief description of the changes here.
 **Checklist before requesting review:**
 
 - [ ] Matches design:
-    * component sizes, spacing, and styles
-    * font size, weight, colour
-    * spelling has been double checked
+
```
d8660091: This is fix by pre-commit script run by husky

client/uikit/PageLayout/index.js
```diff
@@ -3,9 +3,9 @@ import styled from '@emotion/styled';
 import Container from '../Container';
 
 export const PageContainer = styled('div')`
-  min-height: 100%;
-  display: flex;
-  flex-direction: column;
+  display: grid;
```
d8660091: Changed to grid because this is a bug of flexbox
"when the form gets longer, if errors, the left sidebar doesn't go to the bottom of the page.."

https://github.com/icgc-argo/argo-platform/issues/130#issuecomment-512838466

client/uikit/PageLayout/index.js
```diff
@@ -18,13 +18,12 @@ export const Panel = styled('div')`
 `;
 
 export const PageBody = styled('div')`
+  display: grid;
+  grid-template-columns: ${({ noSidebar }) => (noSidebar ? '1fr' : '304px 1fr')};
   flex: 1;
```
joneubank: If you're switching to grid instead of flex (a good call here) can we remove this flex value?

d8660091: Sure

Issue #149, opened 2019-06-20
Toast when a program is created


rosibaj:
Create a Toast Manager 

- close after a period of time
- global array of toast data
- toasts the toast
- Use the toaster to show the create programs success message

Implement the toast manager in the context of the **program creation**
- [ ] User clicks create program 
- [ ] The program is created. 
- [ ] The user is navigated to the program list page, and the newly created program is in the list.
- [ ] A Success toast appears at the top of the page that says:
`<Program.ShortName> was successfully created!`, where the program shortname is dynamically defined to be the program that was just created.
![Screen Shot 2019-07-30 at 11 26 56 AM](https://user-images.githubusercontent.com/18424138/62144367-aba92f00-b2bf-11e9-87c0-4a0db65282cb.png)

**Program edit** (doesn't navigate away from the page):
- [ ] If the user makes a change, and saves it - there is no indication that it has worked. Expected behaviour: 
- User makes change
- Presses Save button
- Save button disables and page scrolls up (or reloads the page)
- Success toast appears that indicates changes have been saved. "Your changes have been saved."
<img width="449" alt="Screen Shot 2019-07-18 at 10 52 01 AM" src="https://user-images.githubusercontent.com/18424138/61467737-24b49800-a94a-11e9-8218-b53f6b4e60fb.png">

rosibaj:
Testing results:
* Created ROSI-CA and ROSY-CA.  As far as i can tell, they created correctly, but did not get the correct toast.  Will add more details once auth deplopy is worked out.

hlminh2000:
@rosibaj back-end is fixed so looks like this is something wrong with the UI. I'm picking this up again :)

rosibaj:
Cannot successfully create a program.

Video of Issue:
https://drive.google.com/file/d/1wtwV_TITJtqm9rbqN2HxyBqDMRwF0-wO/view
![image (5)](https://user-images.githubusercontent.com/15824252/63614541-535d0880-c5b1-11e9-8c4c-3fcc77cebdc9.png)


@hlminh2000 I cannot have this pass on QA at the moment bc i cannot actually see a program successfully created (issues above)

Issue #251, opened 2019-07-12
Standardize all forms - validation


kcullion:
Generalize all forms so the validations are handled in the same way and the error messages are consistent:

- [ ] All multiselect fields should be using the lg size (see https://github.com/icgc-argo/argo-platform/issues/84)

**Error message format:**

1. For required fields:
- [x] "Please fill in [Field Name]" eg. "Please fill in First Name"

2. For a dropdown, eg Membership Type - if the user doesn't fill it in, the error message is "Invalid Membership Type provided", but they didn't even provide one. So this should use the same required field format:
- [x] Please fill in Membership Type

3. For multiselect:
- [x]  [Field Name] field must have at least 1 item eg "Cancer Types field must have at least 1 item"
(NOTE - this is currently almost correct, just remove the "s" from items)

4. Standardize error messages!

PR #322, opened 2019-07-22
Feat/#60 modal interaction

ciaranschutte:
**Description of changes**

- Edit User Modal static UI and data hookup
- Broke program manage page Users and Profile tabs out into their own components

Please note:
- success toast still needs to be added, preferably after #149 
- gateway queries still need to be verified

**Type of Change**
- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches deisgn:
* font size, weight, colour

### DIFF
--- a/client/components/pages/submission-system/program-management/Users.js
+++ b/client/components/pages/submission-system/program-management/Users.js
@@ -75,6 +75,10 @@ const Users = ({ users, shortName }) => {
         onUserDeleteClick={console.log}
         onUserEditClick={console.log}
         onUserResendInviteClick={handleResendEmailClick}
+        onUserEditClick={({ user }) => {
+          dispatch({ showModal: true, user });
+        }}
+        onUserResendInviteClick={console.log}
       />
       {isResendEmailModalOpen && (
         <ResendEmailModal
--- a/client/components/pages/submission-system/program-management/Users.js
+++ b/client/components/pages/submission-system/program-management/Users.js
@@ -6,6 +6,8 @@ import Toast from 'uikit/notifications/Toast';
 import Portal from 'uikit/Portal';
 import Fade from 'uikit/transitions/Fade';
 import Modal from 'uikit/Modal';
+import EditUserModal from '../modals/editUser';
+import { ModalPortal } from '../layout';
 
 function ResendEmailModal({ user, ...otherProps }) {
   return (
@@ -48,6 +50,15 @@ const Users = ({ users }) => {
       setIsToastOpen(false);
     }
   };
+  const initialState = { selectedUser: null, showModal: false };
+  const reducer = (state, action) => {
+    console.log('reducer', state, action);
+    return action.showModal
+      ? { selectedUser: action.user, showModal: true }
+      : { selectedUser: null, showModal: false };
+  };
+  const [state, dispatch] = React.useReducer(reducer, initialState);
+  const { showModal, selectedUser } = state;
 
   return (
     <div>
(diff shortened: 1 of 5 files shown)


client/components/pages/submission-system/program-management/Users.js
```diff
@@ -0,0 +1,51 @@
...
+import { ModalPortal } from '../layout';
+import { useSubmitFormHook, createUserInput } from './';
+import EDIT_USER_MUTATION from './EDIT_USER_MUTATION.gql';
+
+const Users = ({ users, shortName }) => {
+  const initialState = { selectedUser: null, showModal: false };
+  const reducer = (state, action) =>
+    action.showModal
+      ? { selectedUser: action.user, showModal: true }
+      : { selectedUser: null, showModal: false };
```
hlminh2000: If we use reducers, I recommend designing actions in the shape `{type: string, payload: any}`, then switching on the `type` field to return new state.

Though reducers are probably more useful for dealing with a very large state object that needs very strict control over the available mutations to it, and/or when there's a need for capturing mutations in serializable format, so it's a little overkill here and is adding quite a bit of noise. 

It actually looks like `!!selectedUser` seems like the only check we need to decide whether to show the modal, so the `showModal` state isn't needed either. I think we can name `selectedUser` something more like: `currentEdittingUser`, then just check on `!!currentEdittingUser` for the modal. Then later we can have a `currentDeleteingUser` then we can do the same thing to show a different modal. Can be easily done with two `useState` :)

What do you think?

client/components/pages/submission-system/program-management/index.js
```diff
@@ -121,7 +113,7 @@ export default ({ logOut, pathname }: { logOut: any => any, pathname: string })
               )}
             </Tab>
           </Tabs>
-          {activeTab === TABS.USERS && <Users users={FAKE_USERS} />}
+          {activeTab === TABS.USERS && <Users shortName={shortName} users={FAKE_USERS} />}
```
hlminh2000: maybe `shortName` prop should be `programShortName`. The current name makes sense for us now but it kinda reads like `shortName` is some kind of user identifier, which can get a bit confusing in future. Inside the `Users` component, the name `shortName` also loses the context that it's a program identifier.

client/components/pages/submission-system/program-management/index.js
```diff
@@ -1,34 +1,34 @@
...
 };
 
+export const createUserInput = ({ data, shortName }) => ({
+  programShortName: shortName,
+  userFirstName: data.firstName,
+  userLastName: data.lastName,
+  userEmail: data.email,
+  userRole: data.role,
+});
+
```
ciaranschutte: will move these to common file in standardise form ticket: #251

client/components/pages/submission-system/program-management/index.js
```diff
@@ -1,34 +1,34 @@
...
+import Profile from './Profile';
 
-const useSubmitFormHook = ({ gql }) => {
+export const useSubmitFormHook = ({ gql }) => {
   const [triggerMutation, rest] = useMutation(gql);
 
   return [triggerMutation];
 };
 
+export const createUserInput = ({ data, shortName }) => ({
```
hlminh2000: let's call `shortName` `programShortName`

client/components/pages/submission-system/program-management/Users.js
ciaranschutte: Yup sounds good.

Removing the redundant state of showModal actually removes the need for the reducer anyway.
I was using it to change both state at the same time.

ciaranschutte:
@hlminh2000 addressed your feedback
Merged in Xu's resent modal which has a lot of state in it like I had before eg.
```
  const handleResendEmailClick = ({ user }) => {
    setUser(user);
    setIsResendEmailModalOpen(true);
  };
```

I can address this in another PR?

ciaranschutte:
@hlminh2000 sounds good (that's why I left it out of my Edit Modal code for now) :)

PR #325, opened 2019-07-22
added some aria attributes

ciaranschutte:
**Description of changes**

added missing aria roles

for some reason storybook still complains about no aria-checked on Radio. but it's wrong.
aria-wrong

**Type of Change**
- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches deisgn:
* font size, weight, colour

### DIFF
--- a/client/.flowconfig
+++ b/client/.flowconfig
@@ -1,6 +1,4 @@
 [ignore]
-.*/node_modules/*
-
 [include]
 
 [libs]
--- a/client/uikit/form/Checkbox/index.js
+++ b/client/uikit/form/Checkbox/index.js
@@ -106,6 +106,11 @@ const Checkbox = ({ checked, disabled, onChange, ...props }) => {
   );
 };
 
-Checkbox.propTypes = {};
+Checkbox.propTypes = {
+  ['aria-label']: PropTypes.string.isRequired,
+  checked: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
+  onChange: PropTypes.func.isRequired,
+};
 
 export default Checkbox;
--- a/client/uikit/form/Checkbox/index.js
+++ b/client/uikit/form/Checkbox/index.js
@@ -1,5 +1,7 @@
+// @flow
 import React from 'react';
 import PropTypes from 'prop-types';
+// $FlowFixMe no flow-typed definition
 import styled from '@emotion/styled';
 
 /**
@@ -81,7 +83,18 @@ export const StyledCheckbox = styled('div')`
 /*
  * Basic checkbox input
  */
-const Checkbox = ({ checked, disabled = false, onChange, ...props }) => {
+const Checkbox = ({
+  checked,
+  disabled = false,
+  onChange,
+  'aria-label': ariaLabel,
+  ...props
+}: {
+  checked: boolean,
+  disabled: boolean,
+  onChange: (e: any | void) => any | void,
+  'aria-label': string,
+}) => {
   const HiddenCheckboxRef = React.createRef();
 
   return (
@@ -92,12 +105,12 @@ const Checkbox = ({ checked, disabled = false, onChange, ...props }) => {
         checked={checked}
         disabled={disabled}
         onChange={onChange}
-        aria-label={props['aria-label']}
+        aria-label={ariaLabel}
       />
       <div
         className="checkbox"
         onClick={e => {
-          if (document.activeElement !== HiddenCheckboxRef.current) {
+          if (document.activeElement !== HiddenCheckboxRef.current && HiddenCheckboxRef.current) {
             HiddenCheckboxRef.current.focus();
           }
         }}
--- a/client/uikit/form/Checkbox/index.js
+++ b/client/uikit/form/Checkbox/index.js
@@ -1,7 +1,6 @@
 // @flow
 import React from 'react';
 import PropTypes from 'prop-types';
-// $FlowFixMe no flow-typed definition
 import styled from '@emotion/styled';
 
 /**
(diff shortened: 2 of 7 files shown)


hlminh2000:
does it maybe wants it as a string?

ciaranschutte:
@hlminh2000 no dice :(

hlminh2000:
just noting our side conversation, we found that it wants the `aria-checked` property on the containing div as well :)

client/uikit/form/Checkbox/index.js
```diff
@@ -81,7 +81,7 @@ export const StyledCheckbox = styled('div')`
 /*
  * Basic checkbox input
  */
-const Checkbox = ({ checked, disabled, onChange }) => {
+const Checkbox = ({ checked, disabled, onChange, ...props }) => {
```
joneubank: Why not add ariaLabel as a named input and add all these 4 to the propTypes expected by the component?

client/uikit/form/Radio/index.js
```diff
@@ -72,17 +72,22 @@ export const StyledRadio = styled('div')`
  * ::before - checked dot
  * ::after - circle outline
  */
-const Radio = ({ value, disabled, checked, onChange }) => {
+const Radio = ({ value, disabled, checked, onChange, ...props }) => {
```
joneubank: Same propTypes comment

client/uikit/form/Checkbox/index.js
hlminh2000: because I am terrible and have used `aria-label` instead of `ariaLabel`, and that's not a valid name. The good thing is that matches with the Dom's prop name

joneubank: True... but if its a declared prop, and required, we could make our tools check for any form elements that forget to include this property instead of relying on accessibility tests to find missing aria-labels...

As a dev, adding `ariaLabel="whatever"` feels natural on a React component

hlminh2000: it should be in propTypes, can do something like:
```
Checkbox.propTypes = {
  ["aria-label"]: PropTypes.string
}
```
Or use flow

joneubank: Ok sweet. I'm into that.

client/uikit/form/Checkbox/index.js
```diff
@@ -105,6 +106,11 @@ const Checkbox = ({ checked, disabled, onChange }) => {
   );
 };
 
-Checkbox.propTypes = {};
+Checkbox.propTypes = {
+  ['aria-label']: PropTypes.string.isRequired,
+  checked: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
+  onChange: PropTypes.func.isRequired,
```
hlminh2000: I don't think `disabled` really needs to be required eh? can be false by default, not every checkbox needs to configure this :)

client/uikit/form/Radio/index.js
```diff
@@ -97,6 +102,11 @@ const Radio = ({ value, disabled, checked, onChange }) => {
   );
 };
 
-Radio.propTypes = {};
+Radio.propTypes = {
+  ['aria-label']: PropTypes.string.isRequired,
+  checked: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
```
hlminh2000: .

client/uikit/form/Checkbox/index.js
```diff
@@ -92,6 +92,7 @@ const Checkbox = ({ checked, disabled, onChange }) => {
         checked={checked}
         disabled={disabled}
         onChange={onChange}
+        aria-label={props['aria-label']}
```
joneubank: does this put aria-label="undefined" if none is provided?

client/uikit/form/Checkbox/index.js
```diff
@@ -105,6 +106,11 @@ const Checkbox = ({ checked, disabled, onChange }) => {
   );
 };
 
-Checkbox.propTypes = {};
+Checkbox.propTypes = {
```
joneubank: Following @hlminh2000 's  advice to me... Should we use flow type declaration in the Checkbox definition so that our checker catches errors?

client/uikit/form/Radio/index.js
```diff
@@ -97,6 +102,11 @@ const Radio = ({ value, disabled, checked, onChange }) => {
   );
 };
 
-Radio.propTypes = {};
+Radio.propTypes = {
```
joneubank: Same propTypes -> flow types comment.

client/uikit/form/Checkbox/index.js
hlminh2000: Might need to check storybooks config to make sure flow is configured for uikit, but if it is, sure I think that's better

_Sent from my OnePlus5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

ciaranschutte: gotta love that FastHub

client/uikit/form/Checkbox/index.js
ciaranschutte: Yes. PropType required. changing to flow type that will be enforced.

ciaranschutte:
@joneubank @hlminh2000 
- replaced proptypes with Flow
- changed flow ignore, Flow needs to know about node_modules or imports start throwing errors

storybook works with flow, just add the `// @flow` comment to file

client/.flowconfig
```diff
@@ -1,6 +1,4 @@
 [ignore]
-.*/node_modules/*
-
```
hlminh2000: hmm shouldn't this be ignored?

ciaranschutte: if its not ignored, flow will throw errors on some imports @hlminh2000

PR #327, opened 2019-07-23
add delete user modal

ciaranschutte:
**Description of changes**

Adds remove user UI modal

**Type of Change**

- [ ] Bug
- [x] Styling
- [x] New Feature

- [x] Matches deisgn:

* font size, weight, colour

### DIFF
--- a/client/components/pages/submission-system/modals/deleteUser/index.js
+++ b/client/components/pages/submission-system/modals/deleteUser/index.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import Modal from 'uikit/Modal';
+
+const DeleteUserModal = ({ user, dismissModal, onSubmit }) => {
+  const submitForm = async () => {
+    try {
+      const result = onSubmit();
+    } catch (err) {
+      console.log(err);
+    }
+  };
+
+  return (
+    <Modal
+      title="Remove User?"
+      titleIconConfig={{ name: 'warning', fill: 'warning' }}
+      actionButtonText="Save"
+      cancelText="Cancel"
+      onActionClick={() => submitForm()}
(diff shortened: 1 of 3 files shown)


joneubank:
My number one request is that your checkboxes in the PR are formatted correctly 😛

ciaranschutte:
> My number one request is that your checkboxes in the PR are formatted correctly

gosh darn it markdown -_-

client/components/pages/submission-system/modals/deleteUser/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Modal from 'uikit/Modal';
+
+const DeleteUserModal = ({ user, dismissModal, onSubmit }) => {
+  const submitForm = async () => {
+    try {
+      const result = onSubmit();
```
hlminh2000: what would cause this to fail? 
if it's an async operation, I think we're looking for `await onSubmit()`. 
Though I don't know if this is where we should be handling error, unless there's a `disabled` or `loading` state of some sore for the modal while the async stuff is going on. If that's not the case, let's just call `onSubmit` :)

client/components/pages/submission-system/modals/deleteUser/index.stories.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import DeleteUserModal from '.';
+
+const DeleteUserModalStory = storiesOf(`${__dirname}`, module).add('Basic', () => (
+  <DeleteUserModal />
```
hlminh2000: maybe let's add `action` for the callbacks :)

PR #328, opened 2019-07-23
Feat/#251 standardise form error

ciaranschutte:
**Description of changes**

Adds two standard error message functions for validations and update current validations to use them

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches deisgn:

* font size, weight, colour

### DIFF
--- a/client/global/utils/form.js
+++ b/client/global/utils/form.js
@@ -1,3 +1,4 @@
 // error messages
 export const requiredError = fieldName => `Please fill in ${fieldName}.`;
-export const multiSelectError = fieldName => `${fieldName} field must have at least 1 item.`;
+export const mustHaveMoreThanZeroError = fieldName =>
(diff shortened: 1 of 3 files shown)


client/global/utils/form.js
```diff
@@ -0,0 +1,3 @@
+// error messages
+export const requiredError = fieldName => `Please fill in ${fieldName}.`;
+export const multiSelectError = fieldName => `${fieldName} field must have at least 1 item.`;
```
hlminh2000: what do you think of `mustHaveOneError`

ciaranschutte: @hlminh2000 changed to `mustHaveMoreThanZeroError`
quite verbose but matches yups `moreThan` phrasing
and `mustHaveOneError` implies that the field must have exactly one

PR #398, opened 2019-08-01
Feat/#307 form consolidate

ciaranschutte:
**Description of changes**

Make all forms use 'useFormHook'
Remove "sections" functionality; add user modal now uses a map to keep track of individual forms

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


client/components/pages/submission-system/create-program/CreateProgramForm.js
```diff
@@ -105,69 +92,57 @@ const createUpdateProgramInput = formData => ({
...
-  const [validationErrors, setValidationErrors] = React.useState({});
+  const { errors, data, setData, validateField, validateForm, touched, hasErrors } = useFormHook({
+    initialFields: formData,
+    schema: programSchema,
+  });
 
-  const programSchema = isEditing ? updateProgramSchema : createProgramSchema;
+  // TODO: ugly as sin
+  const validationErrors = errors[0];
+  const form = data[0];
```
ciaranschutte: @hlminh2000 @joneubank any thoughts on a better way to do this?
Use case of having the forms in the format of `[{name: 'John', age: 45}, {name: 'Pat', age: 54}]` is `AddUser` needs multiple sections of users to all be the same overall form data.

ciaranschutte:
@hlminh2000 removed "sections" of form hook
used the individual form map as you suggested

@joneubank @hlminh2000 How do we want to merge this because there could be breaking changes? Should we wait until the form integration tests are merged and then we can make sure it's working as it should?

joneubank:
RE: merging... i'm just going to run it locally and check the create and edit program form variants.

client/global/hooks/useFormHook.js
```diff
@@ -0,0 +1,108 @@
...
+};
+
+/**
+ * Supports single forms only, {} not [{},{}]
+ */
+const useFormHook = ({
+  initialFields,
+  schema: formSchema,
+}: {
+  initialFields: {},
```
hlminh2000: should this be `{[k:string]: any}`?

client/global/hooks/useFormHook.js
```diff
@@ -0,0 +1,108 @@
...
+  schema: {
+    validate: typeof yup.object,
+  },
+}) => {
+  let initErrors = {};
```
hlminh2000: should this be `{[k:string]: Error}` or whatever we expect the error type to be?

ciaranschutte: changed to {[k:string: string}
Error is some type from node

ciaranschutte:
> if it's not too much work, can we try flowing the `components/pages/submission-system/create-program/CreateProgramForm.js ` file? :) Don't want to block development though, so can go in separate PR if we do it :)

i'll do this in separate PR

joneubank:
Form is submitting same as before now using the form hook, but just as before it always shows an error toast (even on success). Do you want to look into why or open this as a separate bug? I'm good with either.

hlminh2000:
> Form is submitting same as before now using the form hook, but just as before it always shows an error toast (even on success). Do you want to look into why or open this as a separate bug? I'm good with either.

I think I should look into that as it was my ticket originally to do the toast :)

PR #399, opened 2019-08-01
Add endpoint to retrieve invitation

d8660091:


server/schemas/Program/index.js
```diff
@@ -29,6 +41,16 @@ const typeDefs = gql`
     EXPIRED
   }
 
+  type JoinProgramInvite {
+    id: String!
+    createdAt: Timestamp!
+    expiresAt: Timestamp!
```
joneubank: Can we switch `createdAt` and `expiresAt` to DateTime type borrowed from this library? https://www.npmjs.com/package/graphql-iso-date#datetime

joneubank: If you feel like writing the scalar type with resolvers yourself instead of bringing in the dependency that would be fine to. I'd still name it `DateTime`

server/schemas/Program/index.js
```diff
@@ -29,6 +41,16 @@ const typeDefs = gql`
     EXPIRED
   }
 
+  type JoinProgramInvite {
```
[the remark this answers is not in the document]
d8660091: Are we sure we only have join program invite?

server/schemas/Program/index.js
```diff
@@ -187,6 +214,21 @@ const convertGrpcProgramToGql = programDetails => ({
   membershipType: get(programDetails, 'program.membership_type.value'),
 });
 
+// Convert grpc object to gql object. Make sure your gql definitions matches proto definitions
+function grpcToGql(obj) {
```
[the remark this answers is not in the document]
d8660091: We can move this function to grpc utils. It not only converts keys to CamelCase, but also extract `value`, so convertKeysToCamelCase is not accurate.

server/schemas/Program/index.js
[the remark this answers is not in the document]
joneubank: Ok, good point. Can leave as is.

server/schemas/Program/index.js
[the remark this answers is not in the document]
joneubank: Ah, good point. Leave the name then if you can add a comment just describing what the function does (key names to camelCase and unwrap values).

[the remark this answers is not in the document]
d8660091: will do

server/schemas/Program/index.js
```diff
@@ -205,6 +225,13 @@ const resolvers = {
       return users;
     },
   },
+  DateTime: new GraphQLScalarType({
+    name: 'DateTime',
+    description: 'A string in simplified extended ISO format',
+    serialize(value) {
+      return value;
+    },
+  }),
```
hlminh2000: if we're not using any third-party library, I think this should be defined somewhere else and imported here. Program service and grpc isn't the only use case for this.

Also I think `serialize` should receive a `Date` object and do the serialization to iso string here. That way the contract for future useage is simply to return a `Date` object. A type check with `instanceof Date` should do :)

joneubank: Agreed that this should live in a common schema file, perhaps at `server/schemas/scalars.js`.

server/schemas/Program/index.js
```diff
@@ -29,6 +33,17 @@ const typeDefs = gql`
     EXPIRED
   }
 
+  type JoinProgramInvite {
+    id: String!
```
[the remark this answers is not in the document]
d8660091: right

server/utils/grpcUtils.js
```diff
@@ -89,3 +96,35 @@ export const withRetries = (
...
+  const isTimestamp =
+    keys(maybeTimestamp).length === 2 &&
+    has(maybeTimestamp, 'seconds') &&
+    has(maybeTimestamp, 'nanos');
+  if (isTimestamp) {
+    const { seconds, nanos } = maybeTimestamp;
+    return new Date(seconds * 1000 + nanos / 1000000).toISOString();
+  }
+  return maybeTimestamp;
+}
```
hlminh2000: @joneubank should `grpcToGql` and `timestampToDateTime` really be a grpc thing or are they more a program service thing? the `{ seconds, nanos }` and `{ value }` shape appears to me more like a program service design than grpc, but I'm not quite familiar enough to tell :P

joneubank: Seconds/nanos and value wrapped objects standard-ish grpc things. Conceivably used in any grpc service.

We may only have this one GRPC service now but I will still argue these are generic not ProgramService specific.

hlminh2000: thanks for the clarification! :)

PR #401, opened 2019-08-06
Implements auth context as global context

hlminh2000:
**Description of changes**

Implements the auth context as actual context with initial value set from server-side. This avoids the null initial state that only gets flipped when JS is executed in client, resulting in the flash. Also this fixes https://github.com/icgc-argo/argo-platform/issues/404 which was caused by the local component's auth state not properly flipping.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/client/global/hooks/useAuthContext.js
+++ b/client/global/hooks/useAuthContext.js
@@ -5,29 +5,37 @@ import urlJoin from 'url-join';
 import Cookies from 'js-cookie';
 import Router from 'next/router';
 
-import { decodeToken } from 'global/utils/egoJwt';
+import { decodeToken, isValidJwt } from 'global/utils/egoJwt';
 import { EGO_JWT_KEY } from 'global/constants';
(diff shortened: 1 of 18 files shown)


client/global/hooks/useAuthContext.js
```diff
@@ -69,4 +74,9 @@ export default ({ onError = () => {} }: UseEgoTokenInput = {}) => {
       },
     },
   );
-};
+  return <AuthContext.Provider value={authData}>{children}</AuthContext.Provider>;
+}
+
+export default function useAuthContext() {
+  return React.useContext(AuthContext);
```
d8660091: This function is a little bit trivial, `React.useContext(AuthContext)` looks almost the same as `useAuthContext()`

hlminh2000: lol yeah the name was already there from before so I didn't want to go through and change all the consumption :P
thanks for the review!

PR #406, opened 2019-08-07
Installs Google Analytics

hlminh2000:
**Description of changes**

Initial install of google analytics
- uses [react-ga](https://github.com/react-ga/react-ga)

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


client/pages/_app.js
```diff
@@ -118,6 +119,15 @@ const Root = (() => {
       }
     });
 
+    React.useEffect(() => {
+      ReactGA.initialize(GA_TRACKING_ID, {
+        gaOptions: {
+          userId: data ? data.context.user.email : 'NA',
```
d8660091: maybe `_.get(data, 'context.user.email', 'NA')` is safer

hlminh2000: this is actually very safe because `data` is flow typed against ego's jwt model so any shape mismatch will be caught in build-time. `_.get(data, 'context.user.email', 'NA')` is runtime-safe but will not catch it in build time. Any change to the ego jwt model will be a big breaking change across all services so I don't think we're worried about it changing much :P

joneubank: I half agree with Xu here. We are relying on Ego to provide a token of the correct shape, but build time checks can't ensure this. We need to have run time confirmation of the JWT shape, which should be taken care of in the Auth Context hook (not here). Once that confirmation is done, this code can stay as is.

If that isn't done at run time then use of `_.get` is useful.

joneubank: And on further review this is more or less what we're doing. The expected shape is declared in the file `client/global/utils/egoJwt/index.js` and if the returned shape is different many places will break, so this does not introduce any issues.

As @hlminh2000 pointed out using `_.get` will not let us do flow type checking against our expected shape so it should be avoided in this case.

PR #416, opened 2019-08-08
Hook data to the join program page

d8660091:
Please add a brief description of the changes here.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


client/components/pages/submission-system/register/GET_INVITE.gql
```diff
@@ -0,0 +1,14 @@
+query($inviteId: ID!) {
+  joinProgramInvite(id: $inviteId) {
+    createdAt
+    acceptedAt
```
joneubank: value unused, is it needed?

joneubank: applies to:
- createdAt
- acceptedAt
- emailSent
- program.shortName

Only seeing `user.role` and `program.name` being used.

client/components/pages/submission-system/register/index.js
```diff
@@ -19,26 +15,41 @@ export default ({
...
+}: any) => {
+  const { inviteId } = useRouter().query;
+
+  const { data: { joinProgramInvite } = {}, loading, errors } = useQuery(GET_INVITE, {
+    variables: { inviteId },
+  });
+
+  return (
+    <SubmissionLayout pathname={pathname} logOut={logOut} noSidebar>
+      <div
```
joneubank: Do we have any components that are full page containers like this? Do we use something like this elsewhere? Feels like there iis an opportunity for a reusable UI kit component here.

client/components/pages/submission-system/register/index.js
```diff
@@ -19,26 +15,41 @@ export default ({
...
+          top: 0px;
+          position: absolute;
+        `}
+      >
+        <JoinProgramForm
+          inviteId={inviteId}
+          programName={get(
+            joinProgramInvite,
+            'program.name',
+            loading ? '[loading...]' : '[error to fetch data]',
```
joneubank: Thank you for handling the error case instead of over looking this. 

Instead of this text which is inconsistent with the rest of the application maybe we should replace the entire form with an error page... If we were unable to fetch user role or program name then we should probably just stop the entire register from taking place until we know what went wrong...

Can you get the values outside of the return statement, and if they are missing (error case) we can replace the entire form with an error screen. Should be asking Kim what this looks like.

@hlminh2000 is that an overkill idea? can you help me simplify this?

joneubank: Same comment about the `loading...` text, should we just replace the entire form with a DNASpinner until the data is loaded?

client/components/pages/submission-system/register/joinProgramForm/index.js
```diff
@@ -71,122 +76,209 @@ const ProgramInfoCard = ({ programName, userRole }) => (
...
-                </Col>
-              </Row>
-            </FormControl>
-          </Col>
-        </Row>
+const schema = yup.object().shape({
+  institutions: yup
+    .array()
+    .default([])
+    .required(),
```
joneubank: Needs to provide the standardized error messaging provided in `global/utils/form`.

apples to all required fields:

```
import { requiredError } from 'global/utils/form';
...
yup.string().required(requiredError('Short Name'));
```

client/components/pages/submission-system/register/joinProgramForm/index.js
```diff
@@ -71,122 +76,209 @@ const ProgramInfoCard = ({ programName, userRole }) => (
...
   userRole: string,
-  availableInstitutions: Array<any>,
-  onJoinClick: any => any,
-}) => (
-  <FormContainer>
-    <Row nogutter>
-      <Typography
-        variant="title"
-        color="primary"
+}) => {
```
joneubank: Can this implement the useFormHook for consistency (look at the addUser modal).

client/global/utils/egoJwt/index.js
```diff
@@ -64,8 +62,12 @@ export const canReadProgram = (args: { egoJwt: string, programId: string }): boo
 export const canWriteProgram = (args: { egoJwt: string, programId: string }): boolean =>
   TokenUtils.canWriteProgram(args);
 
-export const isProgramAdmin = (args: { egoJwt: string, programId: string }): boolean =>
-  TokenUtils.isProgramAdmin(args);
+export const isProgramAdmin = (args: { egoJwt: string, programId: string }): boolean => {
+  if (isEmpty(args.egoJwt)) {
+    return false;
```
joneubank: Does the TokenUtils library not check for empty egoJwt?

I'd rather implement this check in that library if it is erroring or failing for the wrong reason. shouldn't need this here.

client/components/pages/submission-system/register/index.js
hlminh2000: I think that's fair. if it fails to find the program, something bigger is likely at play so we can probably just throw a 500 error and let the error page handle it. 
Although, if it fails to find the program because of a known business-case (say, program is deleted / invitation expire, etc... ), we can inform user of the reason. That needs to come from spec though

joneubank: That would be an extremely unusual case, i think a generic "somethign went wrong" message would be fine.

The fields are required for the program and user so we shouldn't ever get them back blank without an error on the request (in theory). So i think just a generic "something went wrong" error screen would be fine.

hlminh2000: then I suggest just throwing a 500 :)

joneubank: use the standard 500 error page?

hlminh2000: no, actually throwing an error and let Next handle showing the error page :P

joneubank: implying that the gql fetch is done serverside? if thats happening then yeah let next handle it, but it wasn't clear to me that this wasnt loading client side.

hlminh2000: it could load either side, if we update the `client/pages/_error.js` to show `<Page500 />` as default for unknown error, that should work both sides :)

joneubank: would fix the other error i brought up this morning too

client/components/pages/submission-system/register/joinProgramForm/index.js
d8660091: How about https://github.com/jquense/yup#using-a-custom-locale-dictionary? By using this function, we don't need to import requiredError and manually set the parameter

joneubank: I like that idea, much more consistent across all forms.

I would want to implement that as a separate PR though, not bundled with this one. Assuming we work to add that in we can leave the error message blank here. I'll add an issue to track that work.

d8660091: thanks!

client/components/pages/submission-system/register/index.js
d8660091: Throw 500 based on your discussion.

hlminh2000: can also just prefetch the data server-side too. check out `client/pages/submission/program/index.js`, the `getGqlQueriesToPrefetch` enables that

hlminh2000: can also just prefetch the data server-side too. check out client/pages/submission/program/index.js, the getGqlQueriesToPrefetch enables that

d8660091: changed to dna spinner for consistency

client/components/pages/submission-system/register/index.js
d8660091: SubmissionLayout is widely used, but I think it's ok to keep it outside of the ui kit simply because its name contains Submission. It's weird for our UI staff to have anything to do with business logic. But we can change the name for sure.

client/global/utils/egoJwt/index.js
hlminh2000: it should, I will update core library to handle. for now, the thin wrapper in this repo has been updated to handle maybe string already so a rebase should do the job @d8660091

d8660091: No, it doesn't. All checking methods assume the egoJWT's type to be string. It's OK to keep this behavior and we make sure the parameter is string when we use those methods. It would be easier with the help of flow-type or typescript

components/pages/submission-system/join/joinProgramForm/index.js
```diff
@@ -0,0 +1,280 @@
...
+   *     joinProgramInput: {
+   *       invitationId: inviteId,
+   *       ...data,
+   *     },
+   *   },
+   * }); */
+
+  const submitForm = async () => {
+    const validData = await validateForm();
+    // TODO
```
joneubank: What is TODO?

components/pages/submission-system/join/joinProgramForm/index.js
```diff
@@ -0,0 +1,280 @@
...
+        </Typography>
+        <Typography variant="label" color="secondary" bold>
+          {userRole}
+        </Typography>
+      </Row>
+    </div>
+  </InfoCard>
+);
+
+const schema = yup.object().shape({
```
joneubank: Can we separate validations to its own file and import here?

components/pages/submission-system/join/joinProgramForm/index.js
```diff
@@ -0,0 +1,280 @@
...
+    schema: schema,
+  });
+
+  const availableInstitutions = [];
+
+  const handleBlur = fieldKey => _ => validateField({ key: fieldKey });
+
+  const handleChange = fieldName => ({ target }) => setData({ key: fieldName, val: target.value });
+
+  /* const [joinProgram] = useMutation(JOIN_PROGRAM, {
```
joneubank: If this is no longer used lets just remove the code completely (and the associated commented out import)

components/pages/submission-system/join/joinProgramForm/index.js
d8660091: it's handling form submission, no longer needed

compose/docker-compose.yaml
```diff
@@ -5,7 +5,7 @@ services:
   # Ego
   #####################################################
   ego-db:
-    image: "${POSTGRES_IMAGE}"
+    image: '${POSTGRES_IMAGE}'
```
joneubank: yaml allows double or single quotes but double seem standard (will be found in all examples) and I see no reason to change this here. Can you update the auto formatter youre using to leave yaml with double quotes?

components/pages/submission-system/join/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+  const { data: { joinProgramInvite } = {}, loading, error } = useQuery(GET_INVITE, {
+    variables: { inviteId },
+  });
+
+  const [joinProgram] = useMutation(JOIN_PROGRAM);
+
+  if (!!error) {
+    error[ERROR_STATUS_KEY] = 500;
+    throw new Error(error);
```
joneubank: 👍 was looking to make sure we handled the case where the invite wasn't found. Error page should be good enough.

components/pages/submission-system/join/index.js
```diff
@@ -0,0 +1,69 @@
+//@flow
+import { get } from 'lodash';
+import { useRouter } from 'next/router';
+// $FlowFixMe
+import { ERROR_STATUS_KEY } from 'pages/_error';
```
hlminh2000: why `$FlowFixMe` here?

components/pages/submission-system/join/joinProgramForm/index.js
```diff
@@ -0,0 +1,250 @@
...
+import { Input } from 'uikit/form';
+import { INPUT_SIZES } from 'uikit/form/common';
+import FormControl from 'uikit/form/FormControl';
+import FormHelperText from 'uikit/form/FormHelperText';
+import InputLabel from 'uikit/form/InputLabel';
+import MultiSelect, { Option } from 'uikit/form/MultiSelect';
+import Typography from 'uikit/Typography';
+import { set } from 'lodash';
+// $FlowFixMe
+import * as yup from 'yup';
```
hlminh2000: `yup` isn't used in this file eh? let's just remove import

components/pages/submission-system/join/joinProgramForm/validation.js
```diff
@@ -0,0 +1,20 @@
+import * as yup from 'yup';
```
hlminh2000: will it break much if we flow check this file?

compose/docker-compose.yaml
d8660091: Nope, it's not my formatter that changes those double quotes. It's `pretty-quick` that do the formatting in the pre-commit hook. It uses prettier config. In theory the double quote should never exist because we have `"singleQuote": true` in package.json

joneubank: Still appears something different in your setup, everyone including you has made commits to this repo without affecting the yaml and flow-type files before... did something change?

joneubank: looks like it only formats when changes are made to a file... thats what changed. We can live with single quotes here since `pretty-quick` will keep it consistent for us from now on. Still find it surprising that prettier config applies to yaml the same as js but thats another issue lol

components/pages/submission-system/join/index.js
d8660091: `Cannot resolve module pages/_error`

compose/docker-compose.yaml
d8660091: Agree. It changes on this pull request because I did a merge with develop branch. Anyway, resolve for now

PR #499, opened 2019-08-21
update url when user navigates program management tabs

hlminh2000:
**Description of changes**

- updates url when user navigates the programs management tabs
- reimplements form reset with a built-in solution to `useFormHook`

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/ProgramForm.js
+++ b/components/pages/submission-system/ProgramForm.js
@@ -70,11 +70,11 @@ const createUpdateProgramInput = formData => ({
  * *************************************** */
 type FormModel = $Call<typeof useFormHook, { initialFields: {}, schema: { validate: () => {} } }>;
 export default function CreateProgramForm({
-  LeftFooterComponent,
+  leftFooterComponent: LeftFooterComponent,
   program = {},
   onSubmit,
 }: {
-  LeftFooterComponent: React.ComponentType<{
+  leftFooterComponent: React.ComponentType<{
     formModel: FormModel,
   }>,
   program?: {
--- a/components/pages/submission-system/ProgramForm.js
+++ b/components/pages/submission-system/ProgramForm.js
@@ -68,12 +68,15 @@ const createUpdateProgramInput = formData => ({
 /* *************************************** *
  * Form data validation
  * *************************************** */
+type FormModel = $Call<typeof useFormHook, { initialFields: {}, schema: { validate: () => {} } }>;
 export default function CreateProgramForm({
-  leftFooterComponent,
+  LeftFooterComponent,
   program = {},
   onSubmit,
 }: {
-  leftFooterComponent: React.Node,
+  LeftFooterComponent: React.ComponentType<{
+    formModel: FormModel,
+  }>,
   program?: {
     name?: string,
     shortName?: string,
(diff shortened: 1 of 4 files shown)


hlminh2000:
flow is actually pretty awesome, caught me a couple bugs here :)

components/pages/submission-system/ProgramForm.js
```diff
@@ -68,12 +68,15 @@ const createUpdateProgramInput = formData => ({
 /* *************************************** *
  * Form data validation
  * *************************************** */
+type FormModel = $Call<typeof useFormHook, { initialFields: {}, schema: { validate: () => {} } }>;
 export default function CreateProgramForm({
-  leftFooterComponent,
+  LeftFooterComponent,
```
d8660091: First letter of component parameter should be in lower case

hlminh2000: I changed it, cause we're doing `<LeftFooterComponent />` now :P

d8660091: Can we do `leftFooterComponent: LeftFooterComponent`? like https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Icon/Icon.js#L57

hlminh2000: why do we prefer lower case? Upper case seems to better communicate that we expect a component rather than an element to me. 
The original naming I think should have been `leftFooterElement`

d8660091: Yes. That's a preference thing. But our personal preference should yield to the community's preference.

hlminh2000: `react-table` uses upper case to denotes component props though, I think upper case is actually more in-line no?

hlminh2000: https://www.npmjs.com/package/react-table
```
<ReactTable
  TableComponent={Component},
  TheadComponent={Component},
  // etc...
  />
```

d8660091: As far as I know the most two popular react framework Material-UI and ant design is using lower case. The closest thing I found in the official react website is render prop, which is lower case. Another thing to notice is that `children` component parameter is lowercase.

hlminh2000: For render props, I think it's ok to use `render` because it's always used as `this.props.render()`, not `<render/>`, in which case it'd be a leaky abstraction even if that was possible. Had the usage was `<Render/>`, it could be argue otherwise.

As for children, again we're never expecting `<children/>`.

It's no big deal to use `leftFooterComponent: LeftFooterComponent` if that's the standard we want to adopt, but I don't personally see the merit of such standard.

d8660091: It's not for personal. It's mainly for easier code reading for others. 

https://github.com/airbnb/javascript/tree/master/react#props

hlminh2000: alright, changed! :) will keep props lower case moving forward. 
Thanks for helping with establishing our standard ;)

d8660091: Thank you!

PR #518, opened 2019-08-27
Refresh token when after joining a program

d8660091:
After join a program, the user has new ego scopes. If the user is
redirected to the program page with the old ego token, he would be
blocked and get a 403 error.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/join/index.js
+++ b/components/pages/submission-system/join/index.js
@@ -34,7 +34,7 @@ export default ({ firstName, lastName, authorizedPrograms = [] }: any) => {
 
   const toaster = useToaster();
 
-  const { refreshToken } = useAuthContext();
+  const { updateToken } = useAuthContext();
 
   const handleSubmit = async validData => {
     try {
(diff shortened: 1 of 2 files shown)


hlminh2000:
@joneubank @ciaranschutte @d8660091 this solution **definitely works great** :) I just wanna leave a note for some exploratory idea.

This implementation bypasses the gateway, so we loose a little bit of that business-logic-agnosticism in the UI. The alternative to this is setting up a graphql subscription to push new tokens from the server-side. This subscription can be bundled into the client side's `AuthProvider` which will free it up from having to fetch new tokens. On server side, we can build ego's token refresh logic into the mutations which update scopes, and push the new token to the UI through an event dispatch. This way any trigger to a mutation that modifies scope will come with a token update on the client side automatically, and in future, any client that consumes our API does not need to worry about manual token refresh.

This would be exploratory ground for us however, so there may be some uncertainties (such as how it would behave with a horizontally scaled gateway service, which is theoretically the same if we use server-push event). I think the potential payoff and gained knowledge is pretty valuable for future, but that said I'm happy to try that out on a separate thread from this specific feature :)

What do you guys think?

PR #538, opened 2019-08-30
Remove express

d8660091:
Remove express to make next run faster

- Remove express.js since we don't need it to handle dynamic route anymore
- Run `next build` when the docker image is built, instead when the image is run
- Use next to handle global environments. i.e. `import 'global/config'` -> `import getConfig from 'next/config'`

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


next.config.js
```diff
@@ -23,13 +22,13 @@ module.exports = withImages({
...
-    AUTH_DISABLED: process.env.AUTH_DISABLED,
-    GA_TRACKING_ID: process.env.GA_TRACKING_ID,
+  publicRuntimeConfig: {
+    GATEWAY_API_ROOT: process.env.GATEWAY_API_ROOT || 'http://localhost:9000',
+    EGO_API_ROOT: process.env.EGO_API_ROOT || '',
+    EGO_CLIENT_ID: process.env.EGO_CLIENT_ID || '',
+    AUTH_DISABLED: process.env.AUTH_DISABLED || false,
+    GA_TRACKING_ID: process.env.GA_TRACKING_ID || '',
+    EGO_URL:
+      process.env.EGO_API_ROOT + `/api/oauth/login/google?client_id=${process.env.EGO_CLIENT_ID}`,
```
andricDu: In some places we are using `url-join`, here it's just a concat. Can we settle on a single way? The original configuration module also used `url-join`.

d8660091: Right

d8660091:
@andricDu Fixed

package.json
```diff
@@ -2,11 +2,8 @@
   "name": "create-next-example-app",
   "version": "1.2.3",
   "scripts": {
-    "dev": "nodemon index.js",
-    "dev:fast": "next",
-    "dev-ui": "node index.js",
+    "dev": "next",
```
joneubank: Can you update this to `next -p 8080` to keep the port the same as before - its helpful since we set all of our local-dev ego applications to redirect to localhost:8080

d8660091: sure

PR #540, opened 2019-08-30
🔥 Husky

ciaranschutte:
Removed Husky pre commit hooks
Run pre commit checks in npm command

why? Sucks to write out whole commit msg and then a flow type fail makes you type it all over

!!!
No local config avaliable for gitmoji
Has to be configured with gitmoji -g
Default behaviour is everything is added to staging which isn't ideal and makes this awesome command kinda a PITA.
!!!

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,5 +1,5 @@
 {
   "hooks": {
-      "pre-commit": "pretty-quick --staged && npm run flow"
+    
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "create-next-example-app",
-  "version": "1.1.2",
+  "version": "1.2.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


joneubank:
how about keep both?

This way people committing without using `npm run commit` still get forced into flow type check, but using the commit script lets you discover that the flow type will fail before you type all the commit messages

I don't want those not using the the script to get around the flow checks.

ciaranschutte:
I'll investigate a fix for gitmoji.
The point of `npm run commit` is to achieve consistency, so using both defeats this.

hlminh2000:
I'm good with removing the flow check on commit hook now, as we have the check built into CI anyway so PRs that break flow won't be merged.

I still like having prettier run on commits written without the script, just because we cannot remove the option of doing so (or is there?). There will be accidental commits written without the script just by muscle memory :P `prettier-quick` only runs on changed files so if we miss the `git commit`, those files won't be processed on future commits written with the script.

Should we just remove the `flow` check but keep `husky` for prettier?

joneubank:
Flow checks have all been removed with the TypeScript implementation. Important to keep consistency between teh commit script and husky since we can't prevent non-script commits. No work required here.

Feel free to reopen or make a new PR if theres something we've missed here.

PR #545, opened 2019-09-04
Move row selections up

d8660091:
**Description of changes**

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/form/Select/index.js
+++ b/uikit/form/Select/index.js
@@ -6,7 +6,13 @@ import { withProps } from 'recompose';
 
 import { StyledInputWrapper, INPUT_SIZES } from '../common';
 import Typography from '../../Typography';
-import { DropdownIcon, OptionsList, Option, HiddenSelect } from './styledComponents';
+import {
+  DropdownIcon,
+  OptionsList,
+  Option,
+  HiddenSelect,
+  POPUP_POSITIONS,
+} from './styledComponents';
 import useTheme from '../../utils/useTheme';
 
 const Select = ({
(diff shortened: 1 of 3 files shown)


uikit/form/Select/index.js
```diff
@@ -132,6 +133,7 @@ Select.propTypes = {
   disabled: PropTypes.bool,
   onChange: PropTypes.func.isRequired,
   onBlur: PropTypes.func,
+  popupPosition: PropTypes.oneOf(['TOP', 'DOWN']),
 };
```
ciaranschutte: export const popupPositions = {UP: "UP", DOWN: "DOWN"};

uikit/form/Select/styledComponents.js
```diff
@@ -26,8 +26,15 @@ export const OptionsList = styled('ol')`
...
-  top: 100%;
   z-index: 100;
+
+  &.TOP {
+    top: 0;
+    transform: translateY(-100%);
+  }
+  &.DOWN {
+    top: 100%;
+  }
```
ciaranschutte:
```
top: ${({position}) => position === popPositions.UP ? 0 : '100%'
transform:  ${({position}) => position === popPositions.UP ? translateY(-100%) : 'inherit'
```

uikit/form/Select/index.js
d8660091: We have been doing a hack of using runtime variables and module export to do type checking at compiling time. I think we can use Typescript in the future. For now, I'll change it to exported constants as a tradition

d8660091:
@ciaranschutte Will do

uikit/form/Select/index.js
ciaranschutte: Sorry bad position of comment.

I just meant add a constant into this file so it's a fixed value and we can use popPositions.UP instead of "UP" or "TOP"...

PR #557, opened 2019-09-09
Lazy load program list

d8660091:
Program list is slow to fetch. Instead of doing it on server end, I make
the change to do it on client side. So the user can be redirected
immediately.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/programs/index.js
+++ b/components/pages/submission-system/programs/index.js
@@ -97,31 +97,16 @@ export default ({ authorizedPrograms = [] }: any) => {
           padding-top: 0px;
         `}
       >
-        {loading ? (
-          <div
-            css={css`
-              padding: 40px;
-              display: flex;
-              justify-content: center;
-              align-items: center;
-            `}
-          >
-            <DnaLoader />
-          </div>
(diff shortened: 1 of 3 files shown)


components/pages/submission-system/programs/index.js
```diff
@@ -98,16 +97,31 @@ export default ({ authorizedPrograms = [] }: any) => {
...
-          {programs.length} results
-          {/*   <TableFilterInput /> */}
-        </TableActionBar>
-        <ProgramsTable
-          loading={loading}
-          programs={sortedPrograms}
-          onProgramUsersClick={handleProgramUsersClick}
-          onProgramEditClick={handleProgramProfileClick}
-        />
+        {loading ? (
```
joneubank: Why do we need to put this loader? Doesn't the programs table already takes the loading parameter so it will show the dnaloader as the data is coming in?

d8660091: Correct, I'm switching to using table's loader

PR #610, opened 2019-09-16
Bug/manage program users tooltip

hlminh2000:
**Description of changes**

- make Tooltip unmount html when hidden
- adds style to support unmounting (needs to be global)
  - enables importing css files
- make actions tooltip appear on left

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/Tooltip/index.css
+++ b/uikit/Tooltip/index.css
@@ -1,0 +1,6 @@
+/* 
+  We need this file applied globally for react-tippy to work, hence the css file
+ */
+.tippy-popper .leave {
+  opacity: 0;
+}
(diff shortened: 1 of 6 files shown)


uikit/Tooltip/index.css
```diff
@@ -0,0 +1,6 @@
+/* 
+  We need this file applied globally for react-tippy to work, hence the css file
+ */
+.tippy-popper .leave {
```
d8660091: Can we put them in the style block of ApplicationRoot?

hlminh2000: it seems more of a concern for the Tooltip component than the application so I figured it makes more sense here :P do you think we prefer application root? 

If there was another way to make the style apply globally without using css imports while keeping the definition scoped in this component context I'd prefer that too :)

hlminh2000:
@joneubank no more hack to monitor thanks to @d8660091 :D

d8660091:
Great job!

PR #639, opened 2019-09-19
Persist state between page navigations

d8660091:
- Add persistentContext, which contains an usePersistentState hook
that's pretty similar to React.useState except the state is persistent
across page navigations
- Persistent search string in the sidemenu
- Persistent programs in the sidemenu
- Persistent scrolling position of sidemenu

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/ApplicationRoot.js
+++ b/components/ApplicationRoot.js
@@ -4,6 +4,7 @@ import { createHttpLink } from 'apollo-link-http';
 import { ToasterContext, useToastState } from 'global/hooks/toaster';
 import { AuthProvider } from 'global/hooks/useAuthContext';
 import { PageContext } from 'global/hooks/usePageContext';
+import { PersistentContext } from 'global/hooks/usePersistentContext';
 import createInMemoryCache from 'global/utils/createInMemoryCache';
 import getConfig from 'next/config';
 import * as React from 'react';
@@ -14,6 +15,7 @@ import ToastStack from 'uikit/notifications/ToastStack';
 import urljoin from 'url-join';
 import Head from 'components/Head';
 import { ApolloProvider } from '@apollo/react-hooks';
+import get from 'lodash/get';
 
 import type { ClientSideGetInitialPropsContext } from 'global/utils/pages/types';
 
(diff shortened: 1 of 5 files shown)


components/ApplicationRoot.js
```diff
@@ -86,36 +88,47 @@ const ToastProvider = ({ children }) => {
   );
 };
 
-export default function ApplicationRoot({
-  egoJwt,
-  apolloCache,
-  pageContext,
-  children,
-}: {
+export default class ApplicationRoot extends React.Component<{
```
hlminh2000: instead of converting this into a class, we can just make a thin `PersistentStateProvider` component and nest it at the appropriate level in the original component.

Something like:
```
const PersistentStateProvider = ({ children }) => {
  const [persistentContext, setPersistentContext] = React.useState({});

  const setItem = (k, v) => {
    setPersistentContext({
      ...persistentContext,
      [k]: v,
    });
  };
  const getItem = (k, defaultValue) => {
    return get(persistentContext, k, defaultValue);
  };

  return (
    <PersistentContext.Provider value={{ getItem, setItem }}>{children}</PersistentContext.Provider>
  );
};
```

components/pages/submission-system/SideMenu.js
```diff
@@ -117,7 +119,7 @@ const LinksToProgram = (props: { program: SideMenuProgram, isCurrentlyViewed: bo
 };
 
 const MultiProgramsSection = ({ programs }: { programs: Array<SideMenuProgram> }) => {
-  const [programNameSearch, setProgramNameSearch] = React.useState('');
+  const [programNameSearch, setProgramNameSearch] = usePersistentState('programNameSearch', '');
```
hlminh2000: this is pretty easy to get key collision and errors from typos, should we have these keys in a constant and/or statically type the type of our global state object?

d8660091: Currently, the use case is small that it's used only for side menu. I think in the future, the use case of persisting state will still be small. That collision will probably happen but in very small chance

components/ApplicationRoot.js
d8660091: I like and tried this approach. It wasn't working. So I switched to Class relentlessly. But as you pointed out, I gave it another try. And it works. Thanks for catching this issue. That's really helpful!

d8660091:
@hlminh2000 Fixed

PR #680, opened 2019-09-27
Feat/#445 upload sample file reg

ciaranschutte:
**Description of changes**

- change apollo client to use upload link for file upload
- show upload errors
- use a shared fragment with queries to keep consistent
- add script and npm command to generate gql fragment matcher json
- add gql loader to storybook webpack

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

**Checklist before requesting review:**

- [ ] Matches design:

  - component sizes, spacing, and styles
  - font size, weight, colour
  - spelling has been double checked

- [ ] New uikit components have Storybook stories
- [ ] Feature is minimally responsive.
- [x] Manual testing of UI feature.
- [ ] Selenium test is completed and passing.

PR #684, opened 2019-09-27
Feat/#613 sample reg ui updates

ciaranschutte:
**Description of changes**
! need to merge first: https://github.com/icgc-argo/platform-ui/pull/680

- Change the "New" star colour from green to purple. Zeplin: https://zpl.io/VDW9Woq
-  font sizes. the instructions and table legend/file info should be using the "data" font type (12px).
- cut the file name off after 40 characters and append a "..." 
- minimize the space between the legend on the left and the file info on the right to 50px
- remove fixed width on instructions buttons

https://app.zenhub.com/workspaces/argo-5c8c8637777e5f767e486217/issues/icgc-argo/platform-ui/613

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [x] Matches design:

  - font size, weight, colour


components/pages/submission-system/program-sample-registration/FileTable/index.tsx
```diff
@@ -37,27 +34,33 @@ const StarIcon = (props: { active: boolean; className?: string }) => (
...
+      <Typography variant="data" color="secondary_dark">
+        {formatFileName(fileName)}
+      </Typography>{' '}
+      uploaded on{' '}
+      <Typography variant="data" color="secondary_dark">
+        {format(new Date(createdAt), 'MMMM D, YYYY ')}
+      </Typography>{' '}
+      by{' '}
+      <Typography variant="data" color="secondary_dark">
+        {creator}
```
hlminh2000: potential `cannot read "creator" of undefined` since submissionInfo is optional. 
Turning on strict null check for TS would surface too many of these cases so I left it off for now, but flow was checking that properly

hlminh2000: potential `cannot read "creator" of undefined` here since `submissionInfo` is optional. I didn't enable strict null check for TS cause it would have surfaced more errors than I could fix in one go :P

uikit/notifications/Banner/index.tsx
```diff
@@ -13,6 +13,7 @@ const Banner = ({
   content,
   variant,
   size,
+  ...props
```
hlminh2000: this component was meant specifically to be a subset of `Notification`. Spreading props would defeats that purpose, so if there's something missing that's needed, let's add it explicitly. In this case, looks like a margin is all that's needed, a container div can be an alternative

components/pages/submission-system/program-sample-registration/FileTable/index.tsx
ciaranschutte: although optional, it will only render if we have submission info
```            {submissionInfo && <SubmissionInfoArea submissionInfo={submissionInfo} />}
```

best to have a default anyway in case we have no submission creator field, or createdAt etc
default empty string should suffice

@hlminh2000

components/pages/submission-system/program-sample-registration/Instructions/index.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+      registrationEnabled={boolean('registrationEnabled', false)}
+      onUpload={action('upload')}
+      setRegisterState={() => ''}
+      shortName={'PROGRAM-SHORTNAME'}
+      registrationId={'123456'}
+    />
+  </div>
+));
+
+export default InstructionsStories;
```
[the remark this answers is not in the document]
ciaranschutte: removing story

ciaranschutte:
@hlminh2000 added default vals for submissionInfo, even though there is a null check before even rendering :)

components/pages/submission-system/program-sample-registration/FileTable/index.tsx
hlminh2000: if it should only render if there is submission, it shouldn't be really be optional eh? else why make the promise to handle null case

hlminh2000: oh I see, in `SubmissionInfo` all these fields are required anyway eh

PR #692, opened 2019-09-30
Handle success saving feedback

d8660091:
**Description**
Handle success saving feedback

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/useCommonToasters.tsx
+++ b/components/useCommonToasters.tsx
@@ -11,7 +11,7 @@ export default function useCommonToasters() {
         variant: TOAST_VARIANTS.ERROR,
         content: 'Something went wrong, please try again later or contact us for assistance',
       }),
-    changeSaved: () =>
+    onSave: () =>
       toaster.addToast({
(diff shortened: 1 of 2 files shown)


joneubank:
Please add a description of work done to the PR.

components/useCommonToasters.tsx
```diff
@@ -12,5 +11,12 @@ export default function useCommonToasters() {
         variant: TOAST_VARIANTS.ERROR,
         content: 'Something went wrong, please try again later or contact us for assistance',
       }),
+    changeSaved: () =>
```
joneubank: Let's rename to `onSaved()` to indicate this is an action taken in response to an event.

d8660091: Changed to onSave to align with html convention. But as unknownError doesn't follow the convention. It's not necessary to do

PR #711, opened 2019-10-03
Adjust min-height of the program list container

d8660091:
**Description of changes**

Use container's loading spinner to keep the spinner centered

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/programs/index.tsx
+++ b/components/pages/submission-system/programs/index.tsx
@@ -102,7 +102,9 @@ export default function Programs({ authorizedPrograms = [] }: any) {
         id="programs-list-container"
         css={css`
           padding-top: 0px;
+          min-height: 300px;
         `}
+        loading={loading}
       >
         <TableActionBar>
           {programs.length} results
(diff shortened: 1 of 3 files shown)


components/pages/submission-system/programs/index.tsx
```diff
@@ -102,14 +102,17 @@ export default function Programs({ authorizedPrograms = [] }: any) {
         id="programs-list-container"
         css={css`
           padding-top: 0px;
+          min-height: 300px;
```
hlminh2000: do we maybe want this in every container and override on a per-case basis? @kcullion

kcullion: @hlminh2000 you mean set the min-height to 300px then overwrite for certain containers? 
That's what we thought..I think @d8660091 suggested something different. What do we think is easiest going forward?

hlminh2000: yeah that was my understanding too, and I'm ok with that. I think there's probably more use-case with the min-height than without, so would rather do it at the base component.

hlminh2000: ok a quick discussion with @kcullion and scanning through the ticket, we decided that it's good to apply this at the per-case basis. Sorry I was thrown off by the ticket title :P

PR #725, opened 2019-10-04
Fix input inside button which is not supported by an old Firefox version

d8660091:
**Description of changes**

Firefox version 60 in windows 10 does not support focusing on the input
of an button element

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/SubMenu/styledComponents.tsx
+++ b/uikit/SubMenu/styledComponents.tsx
@@ -2,6 +2,10 @@ import React from 'react';
 import styled from '@emotion/styled';
 import css from '@emotion/css';
 import defaultTheme from 'uikit/theme/defaultTheme';
+import {
+  CreateStyledComponentExtrinsic,
+  CreateStyledComponentIntrinsic,
+} from '@emotion/styled/types';
 
 type StyleCalculationInput = {
   selected?: boolean;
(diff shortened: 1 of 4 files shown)


uikit/SubMenu/styledComponents.tsx
```diff
@@ -97,7 +97,7 @@ export const IconContainer = styled('span')`
   align-items: center;
 `;
 
-export const ContentContainer = styled('button')`
+export const ContentContainer = styled<'button', { as?: string }>('button')`
```
hlminh2000: `keyof HTMLElementTagNameMap` might be a better type for this? :)

d8660091: I'm not sure what you mean. But I've found an alternative that's more efficient

hlminh2000: ah! nice :) I just meant instead of `as: string`, as `keyof HTMLElementTagNameMap` is a bit more restrictive

hlminh2000: `HTMLElementTagNameMap` is a type in react that maps html element tag names to some specific react element types. so `keyof HTMLElementTagNameMap` just gives you an enum of all the html tags that react is aware of :P

d8660091: Good point. I think annotating as is a good thing to do. Instead of HTMLElementTagNameMap. I think it's better to align it with styled's existing tag type. Which is JSX.IntrinsicElements

hlminh2000: I see, that's better! thanks for looking into it! :)

d8660091: Thank you!

PR #752, opened 2019-10-15
Improvements on Clinical submission page

hlminh2000:
- fixes the funky error removal behaviour.
  - uses index as the unique identifier for each error
  - updates apollo cache directly instead of going through a local state (`fileError` is ephemeral so no need to worry about being out-of-sync from server-side)

- Hides the badge for each clinical entity when records count is 0
- Implements the success condition on each entity badge when the submission is validated (see https://github.com/icgc-argo/platform-ui/pull/752/commits/95230bf0bfb14077c170a33244ab04c23aaa54a6)
- Implements the purple star for new records (see https://github.com/icgc-argo/platform-ui/pull/752/commits/74c286c05d5dd8aedc658f7b1b66d0cf8313f8ba)
  - Also updates logic to compare row to stats

### DIFF
--- a/components/pages/submission-system/program-clinical-submission/FilesNavigator/index.tsx
+++ b/components/pages/submission-system/program-clinical-submission/FilesNavigator/index.tsx
@@ -73,13 +73,17 @@ export default ({
               >
                 {fileState.displayName}
               </div>
-              {fileState.status !== 'NONE' && fileState.status !== 'ERROR' && (
-                <VerticalTabs.Tag variant={fileState.status}>
-                  {fileState.recordsCount}
-                </VerticalTabs.Tag>
-              )}
-              {fileState.status === 'ERROR' && (
-                <VerticalTabs.Tag variant="ERROR">!</VerticalTabs.Tag>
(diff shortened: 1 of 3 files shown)


components/pages/submission-system/program-clinical-submission/FilesNavigator/index.tsx
```diff
@@ -73,13 +73,17 @@ export default ({
...
                 {fileState.displayName}
               </div>
-              {fileState.status !== 'NONE' && fileState.status !== 'ERROR' && (
-                <VerticalTabs.Tag variant={fileState.status}>
-                  {fileState.recordsCount}
-                </VerticalTabs.Tag>
-              )}
-              {fileState.status === 'ERROR' && (
-                <VerticalTabs.Tag variant="ERROR">!</VerticalTabs.Tag>
+              {!!fileState.recordsCount && (
```
joneubank: `!!truthy` 😹

joneubank: don't change, written this way its actually more obvious when reading

PR #756, opened 2019-10-15
✨ configures the table in SignOffValidationModal to render summary data

hlminh2000:
**Description of changes**

- configures the table in SignOffValidationModal to render diff data
- adds a `withRowBorder` prop for table
- introduces a variant system for table, starting with two values: `"DEFAULT"` and `"STATIC"`, for quick ways to set common prop defaults.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-clinical-submission/SubmissionSummaryTable.tsx
+++ b/components/pages/submission-system/program-clinical-submission/SubmissionSummaryTable.tsx
@@ -1,0 +1,92 @@
+import * as React from 'react';
+import { GqlClinicalSubmissionData } from './types';
+import { capitalize } from 'global/utils/stringUtils';
+import Table, { TableColumnConfig } from 'uikit/Table';
+import { StatArea } from '../common';
+import { FILE_STATE_COLORS } from './FilesNavigator/FileRecordTable';
+import { useTheme } from 'uikit/ThemeProvider';
+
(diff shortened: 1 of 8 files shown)


components/pages/submission-system/program-clinical-submission/SubmissionSummaryTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+      entity =>
+        ({
+          accessor: entity.clinicalType,
+          Header: capitalize(entity.clinicalType.split('_').join(' ')),
+        } as TableColumnConfig<Entry>),
+    ),
+  ];
+  return (
+    <Table
+      variant="STATIC"
```
ciaranschutte: minor suggestion: use a TABLE_VARIANTS.STATIC here, the same way you have in storybook
You may feel its redundant seen as the type check will throw an error if it's not one of the variants  :)

ciaranschutte: @hlminh2000

hlminh2000: yeah I stopped using the run-time constants cause the type check will catch these in build time... the constants are helpful in storybook cause we often want to iterate through possible values though, the down-side is you kinda have to know about its existence :P

PR #757, opened 2019-10-16
Style/#687 sidebar footer

ciaranschutte:
**Description of changes**

- extend the sidebar to the bottom of the browser
- move the footer so it is below the content area
-  right align the footer links beside the OICR logo

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/Footer.tsx
+++ b/components/Footer.tsx
@@ -2,8 +2,10 @@ import * as React from 'react';
 import Footer from 'uikit/Footer';
 import { css } from 'uikit';
 import { APP_VERSION } from 'global/constants';
+import useTheme from 'uikit/utils/useTheme';
 
 export default function GlobalFooter() {
+  const theme = useTheme();
   return (
     <Footer
       version={APP_VERSION}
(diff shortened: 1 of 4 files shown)


components/Footer.tsx
```diff
@@ -2,17 +2,19 @@ import * as React from 'react';
...
 import { css } from 'uikit';
 import { APP_VERSION } from 'global/constants';
+import useTheme from 'uikit/utils/useTheme';
 
 export default function GlobalFooter() {
+  const theme = useTheme();
   return (
     <Footer
       version={APP_VERSION}
+      className="footer"
```
hlminh2000: would it be better to have this className applied directly in the `Footer` component's `Container` tag? 
I'm thinking something like `<Container className={`footer ${props.className || ""}`} />`. I think this is preferable because the `PageLayout` that expects this classname is in uikit so better to keep it close around so it always applies.

What do you think?

ciaranschutte: @hlminh2000 sounds good

PR #759, opened 2019-10-16
✨ implements the redirect when submission signoff succeeds

hlminh2000:
**Description of changes**

- loading state while sign-off in progress (with 2s fake delay)
- checks on mutation result to decide whether to redirect

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-clinical-submission/gql/UPLOAD_CLINICAL_SUBMISSION.gql
+++ b/components/pages/submission-system/program-clinical-submission/gql/UPLOAD_CLINICAL_SUBMISSION.gql
@@ -1,7 +1,10 @@
 #import "./CLINICAL_SUBMISSION_FRAGMENT.gql"
 
 mutation UPLOAD_CLINICAL_SUBMISSION($programShortName: String!, $files: [Upload!]) {
-  uploadClinicalSubmissions(programShortName: $programShortName, clinicalFiles: $files) {
+  clinicalSubmissions: uploadClinicalSubmissions(
+    programShortName: $programShortName
+    clinicalFiles: $files
+  ) {
     ...ClinicalSubmissionFragment
   }
 }
--- a/components/pages/submission-system/program-clinical-submission/index.tsx
+++ b/components/pages/submission-system/program-clinical-submission/index.tsx
@@ -205,23 +205,22 @@ export default function ProgramClinicalSubmission() {
 
   const handleSignOff = async () => {
     try {
-      await getUserApproval()
-        .catch(() => {
-          console.log('sign off canceled');
-        })
-        .then(async () => {
-          setPageLoaderShown(true);
-          await sleep(2000);
-          const { data: newData } = await signOffSubmission({
-            variables: {
-              programShortName,
-              submissionVersion: data.clinicalSubmissions.version,
-            },
-          });
-          if (newData.clinicalSubmissions.state === null) {
-            router.push(PROGRAM_DASHBOARD_PATH.replace(PROGRAM_SHORT_NAME_PATH, programShortName));
-          }
+      const userDidApprove = await getUserApproval();
+      if (userDidApprove) {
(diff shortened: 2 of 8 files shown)


components/pages/submission-system/program-clinical-submission/index.tsx
```diff
@@ -215,6 +225,11 @@ export default function ProgramClinicalSubmission() {
           />
         </ModalPortal>
       )}
+      {pageLoaderShown && (
+        <ModalPortal>
+          <DnaLoader />
+        </ModalPortal>
+      )}
```
hlminh2000: will be discussing with everyone whether this is a standard busy state we should adopt

components/pages/submission-system/program-clinical-submission/gql/UPLOAD_CLINICAL_SUBMISSION.gql
```diff
@@ -1,7 +1,10 @@
 #import "./CLINICAL_SUBMISSION_FRAGMENT.gql"
 
 mutation UPLOAD_CLINICAL_SUBMISSION($programShortName: String!, $files: [Upload!]) {
-  uploadClinicalSubmissions(programShortName: $programShortName, clinicalFiles: $files) {
+  clinicalSubmissions: uploadClinicalSubmissions(
```
hlminh2000: this ensures query result comes back as `{data: {clinicalSubmissions: {...}}}`

components/pages/submission-system/program-clinical-submission/index.tsx
```diff
@@ -178,19 +183,22 @@ export default function ProgramClinicalSubmission() {
...
-            approveSubmission({
-              variables: {
-                programShortName,
-                submissionVersion: data.clinicalSubmissions.version,
-              },
-            }),
-        )
-        .catch(() => {
-          console.log('sign off canceled');
+      const userDidApprove = await getUserApproval();
```
hlminh2000: updated `getUserApproval` to always resolve a boolean rather than throwing an error because that reads better

components/pages/submission-system/program-clinical-submission/index.tsx
```diff
@@ -178,19 +183,22 @@ export default function ProgramClinicalSubmission() {
...
+      if (userDidApprove) {
+        setPageLoaderShown(true);
+        await sleep(2000);
```
joneubank: Should our standard sleep timers be on a constant or something instead of peppering `2000` around our code base?

components/pages/submission-system/program-clinical-submission/index.tsx
```diff
@@ -178,19 +183,22 @@ export default function ProgramClinicalSubmission() {
...
+          variables: {
+            programShortName,
+            submissionVersion: data.clinicalSubmissions.version,
+          },
         });
+        if (newData.clinicalSubmissions.state === null) {
+          router.push(PROGRAM_DASHBOARD_PATH.replace(PROGRAM_SHORT_NAME_PATH, programShortName));
+        }
+      } else {
+        console.log('sign off canceled');
```
joneubank: Do we want this comment permanently? Maybe we need to set up a logger with log levels, don't need this in prod.

components/pages/submission-system/program-clinical-submission/index.tsx
hlminh2000: good idea, could just be a default value for the argument :) Will update!

components/pages/submission-system/program-clinical-submission/index.tsx
hlminh2000: it probably makes sense to have a logger set up. For this case though we can take it off, now that it's written as a condition check rather than an error catch, there isn't a need for the else block anymore :)

PR #763, opened 2019-10-18
Feat/static global config

hlminh2000:
**Description of changes**

Exports serialized `process.env` values from `global/config` so they can be accessed statically, instead of through next's `getConfig`. Why is this better?
- More type safe: You can't access `process.env.NON_EXISTENT_FIELD` as the build will fail. Also it'll come with proper types. `const {NON_EXISTENT_FIELD} = getConfig(). publicRuntimeConfig` will not fail during build and does not come with static types.
- `getConfig` only works in react components, so static utility functions  do not get access to env unless passed in.
- This repo does not handle any back-end business logic, so there isn't and won't be any secret here, we don't need to worry about separating client vs server-side env vars.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature
- [x] Refactor

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/global/utils/egoJwt/index.tsx
+++ b/global/utils/egoJwt/index.tsx
@@ -3,14 +3,11 @@ import memoize from 'lodash/memoize';
 import getConfig from 'next/config';
 
 import createEgoUtils from '@icgc-argo/ego-token-utils/dist/lib/ego-token-utils';
+import { EGO_PUBLIC_KEY } from 'global/config';
 
 let initializedTokenUtils = null;
 const TokenUtils = () => {
   if (!initializedTokenUtils) {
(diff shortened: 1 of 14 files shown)


global/utils/egoJwt/index.tsx
```diff
@@ -3,18 +3,9 @@ import memoize from 'lodash/memoize';
...
-  if (!initializedTokenUtils) {
-    const EGO_PUBLIC_KEY = (getConfig().publicRuntimeConfig.EGO_PUBLIC_KEY || '').replace(
-      /\\n/g,
-      '\n',
-    );
-    initializedTokenUtils = createEgoUtils(EGO_PUBLIC_KEY);
-  }
-  return initializedTokenUtils;
-};
+const TokenUtils = createEgoUtils(EGO_PUBLIC_KEY);
```
joneubank: 👍

PR #775, opened 2019-10-21
Change join program flow

d8660091:
**Description of changes**
- Two steps of join program
- Show welcome banner when join a program

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


uikit/Tabs/index.tsx
```diff
@@ -85,7 +95,7 @@ function Tabs({ value, onChange, children: childrenProp }) {
 
   return (
     <TabsContext.Provider value={context}>
-      <Container>{children}</Container>
+      <Container className={className}>{children}</Container>
```
hlminh2000: just curious, why should this be removed? seems like a neat feature

uikit/notifications/Banner/index.tsx
```diff
@@ -27,6 +27,7 @@ const Banner = ({
     content={content}
     variant={variant}
     size={size}
+    {...props}
```
hlminh2000: if we do this, let's maybe extend the props type to include `& React.ComponentProps<typeof Notification>` so the api is surfaced to consumer :)

global/constants/pages.tsx
```diff
@@ -12,6 +11,8 @@ export const PROGRAM_MANAGE_PATH = `${SUBMISSION_PATH}/program/${PROGRAM_SHORT_N
 export const PROGRAM_DASHBOARD_PATH = `${SUBMISSION_PATH}/program/${PROGRAM_SHORT_NAME_PATH}/dashboard`;
 export const PROGRAM_SAMPLE_REGISTRATION_PATH = `${SUBMISSION_PATH}/program/${PROGRAM_SHORT_NAME_PATH}/sample-registration`;
 export const PROGRAM_CLINICAL_SUBMISSION_PATH = `${SUBMISSION_PATH}/program/${PROGRAM_SHORT_NAME_PATH}/clinical-submission`;
+export const PROGRAM_JOIN_DETAILS_PATH = `${SUBMISSION_PATH}/program/join/details/[inviteId]`;
```
hlminh2000: it's a bit verbose, but maybe let's export the `[inviteId]` part as a const and reuse in the pages?

uikit/Button/GoogleLogin/index.tsx
```diff
@@ -15,10 +16,20 @@ const StyledLink = styled('a')`
...
+  id?: string;
+  link: string;
+  className?: string;
+  redirectPath?: string;
+}> = ({ link, id, className, redirectPath }) => {
+  const handleClick = () => {
+    if (!!redirectPath) {
+      localStorage.setItem(LOCAL_STORAGE_REDIRECT_KEY, redirectPath);
+    }
+  };
```
hlminh2000: I like the idea of having this logic built-into this button :) only problem is it kind of makes this component app-logic aware, so what do you think of just moving it out of `uikit` and into `components`? 
I think it never belonged here anyway :P

components/pages/submission-system/join/details.tsx
```diff
@@ -54,6 +74,7 @@ export default ({ firstName, lastName, authorizedPrograms = [] }: any) => {
 
       const egoToken = await updateToken(); // Side effect
       if (egoToken) {
+        window.localStorage.setItem('justJoinedProgram', joinProgramInvite.program.shortName);
```
hlminh2000: should `"justJoinedProgram"` be in a constant? it's a magic string used by two different file so seems a bit prone to human errors

hlminh2000:
btw TS is complaining :P

uikit/Tabs/index.tsx
hlminh2000: my bad, it wasn't removed, it was added... sorry brain fart

uikit/Button/GoogleLogin/index.tsx
d8660091: Totally agree

components/pages/submission-system/join/details.tsx
d8660091: Good idea

PR #776, opened 2019-10-21
Register sample stats, use rows instead of cols

ciaranschutte:
…al columns

**Description of changes**

Previously the stat counter was taking each individual change as a variable even if on same row eg.
new specimen [row 1] + new sample id [row 1] = 2 new files
changed to take rows into account

added `isNew` property to display this on FileTable 

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-sample-registration/index.tsx
+++ b/components/pages/submission-system/program-sample-registration/index.tsx
@@ -1,6 +1,7 @@
 import { useMutation, useQuery } from '@apollo/react-hooks';
 import usePageContext from 'global/hooks/usePageContext';
 import get from 'lodash/get';
+import union from 'lodash/union';
 import * as React from 'react';
 import { Row } from 'react-grid-system';
 import { css } from 'uikit';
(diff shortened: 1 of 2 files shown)


components/pages/submission-system/program-sample-registration/index.tsx
```diff
@@ -141,20 +146,25 @@ export default function ProgramIDRegistration() {
...
-      newSamples = { count: 0 },
-      newSpecimens = { count: 0 },
+      alreadyRegistered,
+      newDonors,
+      newSamples,
+      newSpecimens,
     } = clinicalRegistration;
     submissionInfo = { createdAt, creator, fileName };
+
+    newRows = union(newDonors.rows, newSamples.rows, newSpecimens.rows);
```
hlminh2000: can we declare `newRows` here directly as a const? :)

ciaranschutte: @hlminh2000 const is block scoped
need `newRows` later to pass into `recordsToFileTable` for new star highlighing

hlminh2000: oh I misread and thought this was outside the if :P thanks!

PR #809, opened 2019-10-24
♻️ Persists apollo client through navigation

hlminh2000:
**Description of changes**

Persists apollo client in react state so that the cache is retained throughout navigation

**Type of Change**

- [x] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/ApplicationRoot.tsx
+++ b/components/ApplicationRoot.tsx
@@ -107,25 +107,32 @@ function PersistentStateProvider({ children }) {
   );
 }
 
-function MyApolloProvider({ children, cache }) {
+const ApolloClientProvider: React.ComponentType<{ egoJwt: string; apolloCache: any }> = ({
+  children,
+  egoJwt,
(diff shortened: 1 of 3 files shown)


components/ApplicationRoot.tsx
```diff
@@ -107,25 +107,31 @@ function PersistentStateProvider({ children }) {
   );
 }
 
-function MyApolloProvider({ children, cache }) {
+const ApolloClientProvider: React.ComponentType<{ egoJwt: string; apolloCache: any }> = ({
```
joneubank: Should this be in its own file? why is it jammed into the app root file?

hlminh2000: no particular reason why other than this is the only place it's used so figured it's ok to keep it close, I didn't feel the file was big enough (maybe just yet?) lol

PR #818, opened 2019-10-28
Hanlde feedbacks of join program flow

d8660091:
- Set min-height of the form so no werid bumps as typing
- Change text of the step 2 so the user name is shown
- Handle unknown error

**Description of changes**

**Type of Change**

- [x] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/join/JoinProgramLayout.tsx
+++ b/components/pages/submission-system/join/JoinProgramLayout.tsx
@@ -20,7 +20,7 @@ export enum InviteState {
   Accepted,
   Pending,
   Revoked,
-  Loading,
+  UnSet,
 }
 
 export default function JoinProgramLayout({
@@ -37,7 +37,7 @@ export default function JoinProgramLayout({
   notFound: boolean;
 }) {
   const { EGO_URL } = getConfig();
-  let inviteState: InviteState = InviteState.Loading;
+  let inviteState: InviteState = InviteState.UnSet;
 
   if (notFound) {
     inviteState = InviteState.NotFound;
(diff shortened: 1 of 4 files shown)


components/pages/submission-system/join/JoinProgramLayout.tsx
```diff
@@ -185,6 +185,7 @@ export default function JoinProgramLayout({
               <DnaLoader />
             </div>
           )}
+          {inviteState == InviteState.UnSet && !loading && <div>Unknown Error</div>}
```
hlminh2000: this case would be a bug right? Are we actually hitting this case? If so, is it better to use a toast or a banner of some sort to ask user to let us know?

@rosibaj any thoughts? This could be a larger conversation, but just wanted to note this case so we don't forget about it :P

rosibaj: what is the sate here @hlminh2000 ?

d8660091: This is not in design. This morning, program service went down. The frontend didn't handle this situation, so I added this line

rosibaj: thanks @d8660091 - this is important and i think falls in the `we need a general something went wrong error toast` that gets universally used.

hlminh2000: we can skip this check for this PR :) @rosibaj we do we already have a ticket for unknown error handling? :)

rosibaj: nope on todo list!

PR #851, opened 2019-11-05
Feat/#779 reg sample sidebar icons

ciaranschutte:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


ciaranschutte:
@hlminh2000 for the case of displaying error icon when file upload is invalid, I propose we refactor to just add `uploadError` field to `clinicalRegistrationData`. thoughts?
https://app.zenhub.com/workspaces/argo-5c8c8637777e5f767e486217/issues/icgc-argo/platform-ui/779

components/pages/submission-system/SIDE_MENU_SAMPLE_REGISTRATION_STATE.gql
```diff
@@ -0,0 +1,10 @@
+query SIDE_MENU_SAMPLE_REGISTRATION_STATE($programShortName: String!) {
+  clinicalRegistration(shortName: $programShortName) {
+    programShortName
+    createdAt
```
hlminh2000: I don't think `createdAt` is used eh?

components/pages/submission-system/SideMenu.tsx
```diff
@@ -64,6 +74,14 @@ type ClinicalSubmissionQueryResponse = {
     }>;
   };
 };
+
+type SampleRegistrationQueryResponse = {
+  clinicalRegistration: {
+    fileName: string;
```
hlminh2000: should we include `programShortName` here too to reflect the actual gql query? It's not actually read anywhere so I see why we might choose to leave it out, just curious if we have a preference

components/pages/submission-system/SideMenu.tsx
```diff
@@ -100,7 +131,24 @@ const LinksToProgram = (props: { program: SideMenuProgram; isCurrentlyViewed: bo
...
+          content={
+            <div
+              css={css`
+                display: flex;
+                justify-content: space-between;
+                width: 100%;
+                align-items: center;
+                padding-right: 15px;
+              `}
+            >
```
hlminh2000: shouldn't this div be the `StatusMenuItem`? :)

components/pages/submission-system/SideMenu.tsx
ciaranschutte: yeah i see no harm in that. minimises potential confusion

components/pages/submission-system/SideMenu.tsx
ciaranschutte: nice catch! :)

PR #862, opened 2019-11-06
missed set loader false

jaserud:
**Description of changes**

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

### DIFF
--- a/components/pages/submission-system/program-clinical-submission/Header.tsx
+++ b/components/pages/submission-system/program-clinical-submission/Header.tsx
@@ -51,8 +51,6 @@ export default ({
   const { data, updateQuery: updateClinicalSubmissionQuery } = useClinicalSubmissionQuery(
     programShortName,
   );
-  // no submission if no submission version
-  const isSubmissionExist = submissionVersion !== undefined && submissionVersion !== null;
 
   const [reopenSubmission] = useMutation<
     ClinicalSubmissionQueryData,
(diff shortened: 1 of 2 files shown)


components/pages/submission-system/program-clinical-submission/Header.tsx
```diff
@@ -51,6 +51,8 @@ export default ({
   const { data, updateQuery: updateClinicalSubmissionQuery } = useClinicalSubmissionQuery(
     programShortName,
   );
+  // no submission if no submission version
+  const isSubmissionExist = submissionVersion !== undefined && submissionVersion !== null;
```
hlminh2000: sorry for being a grammar nazi... `doesSubmissionExist`? :P also can just be `!!submissionVersion`

components/pages/submission-system/program-clinical-submission/Header.tsx
```diff
@@ -140,22 +142,13 @@ export default ({
...
-    if (!submissionVersion) {
-      toaster.addToast({
-        variant: 'INFO',
-        title: 'No submission',
-        content: `There is no submitted files to clear`,
-      });
-      return;
-    }
     try {
       await clearClinicalSubmission();
```
hlminh2000: can we add a `await sleep()` here? it's too fast so the spinner is just a flash (yea... UX is an irrational art :P)

components/pages/submission-system/program-clinical-submission/Header.tsx
jaserud: `!submissionVersion` right?, wasn't sure you'd be ok with that

PR #879, opened 2019-11-09
✅ integrates query validation test

hlminh2000:
**Description of changes**
- updates `queryValidations.test.tsx` to work with graphql import statements
- adds a CI step to run this test against live environment prior to deployment

### DIFF
--- a/gqlValidation/queryValidations.test.ts
+++ b/gqlValidation/queryValidations.test.ts
@@ -8,11 +8,16 @@ import urlJoin from 'url-join';
 import { ABSOLUTE_ROOT_DIR, compileGqlAst, getGqlFiles } from './utils';
 
 require('dotenv').config();
+const GATEWAY_API_ROOT = process.env.GATEWAY_API_ROOT;
+
+if (!GATEWAY_API_ROOT) {
+  throw new Error(`${GATEWAY_API_ROOT} must be provided`);
+}
 
 const GRAPHQL_URL = urlJoin(process.env.GATEWAY_API_ROOT, 'graphql');
 const validationRules = specifiedRules.filter(rule => rule !== NoUnusedFragmentsRule);
 
-describe('this test', async () => {
+describe('gql validations', async () => {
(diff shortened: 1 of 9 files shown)


gqlValidation/queryValidations.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import { expect } from 'chai';
+import urlJoin from 'url-join';
+import { ABSOLUTE_ROOT_DIR, compileGqlAst, getGqlFiles } from './utils';
+
+require('dotenv').config();
+
+const GRAPHQL_URL = urlJoin(process.env.GATEWAY_API_ROOT, 'graphql');
+const validationRules = specifiedRules.filter(rule => rule !== NoUnusedFragmentsRule);
+
+describe('this test', async () => {
```
ciaranschutte: maybe something a little more descriptive like `describe('gql validations')`?

gqlValidation/queryValidations.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import fetch from 'node-fetch';
+import { validate, specifiedRules, NoUnusedFragmentsRule } from 'graphql/validation';
+import path from 'path';
+import { expect } from 'chai';
+import urlJoin from 'url-join';
+import { ABSOLUTE_ROOT_DIR, compileGqlAst, getGqlFiles } from './utils';
+
+require('dotenv').config();
+
+const GRAPHQL_URL = urlJoin(process.env.GATEWAY_API_ROOT, 'graphql');
```
ciaranschutte: maybe throw a more descriptive error if `GATEWAY_API_ROOT` isn't defined?
Otherwise it will be a regex error from url-join which is a little confusing

something quick like
```
require('dotenv').config();
if (!process.env.GATEWAY_API_ROOT) {
  throw 'GATE WAY API NOT DEFINED';
}
```

hlminh2000: thanks, addressed!

PR #893, opened 2019-11-14
Feat/#108 registration schema updates

ciaranschutte:
**Description of changes**

Update clinical reg data to react to backend changes
No longer inspect type of response, just check if we have errors.

Removed custom fragment matcher logic when creating apollo cache
(left script to generate it if we need it in future)

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


components/pages/submission-system/program-sample-registration/index.tsx
```diff
@@ -139,17 +139,11 @@ export default function ProgramIDRegistration() {
...
   const onUpload = async ({ response, fileName }) => {
     // reset error banner
     setErrorBanner({ visible: false, content: '', title: '' });
-    const { __typename: respType, ...resp } = response;
     // display an error banner or clinical data will update with errors array
-    if (respType === responseTypes.CLINICAL_REG_INVALID) {
+    if (response.fileErrors.length > 0) {
       showError({
         errorCode: ERROR_CODES.INVALID_FILE_NAME.code,
         fileName: formatFileName(fileName),
```
hlminh2000: since the mutation returns the same type as the query now, instead of imperatively calling `showError`, we should be able to just conditionally render the error eh?

PR #898, opened 2019-11-15
🐛 default error table row index should be 1-index

hlminh2000:
**Description of changes**

Just making sure error tables also display rows with 1-index

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/ErrorNotification.tsx
+++ b/components/pages/submission-system/ErrorNotification.tsx
@@ -13,6 +13,7 @@ export const defaultColumns = [
     accessor: 'row',
     Header: 'Row #',
     maxWidth: 70,
+    Cell: ({ original }) => <>{Number(original.row) + 1}</>,
   },
   {
     accessor: 'donorId',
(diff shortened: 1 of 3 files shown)


components/pages/submission-system/ErrorNotification.tsx
```diff
@@ -13,6 +13,7 @@ export const defaultColumns = [
     accessor: 'row',
     Header: 'Row #',
     maxWidth: 70,
+    Cell: ({ original }) => <>{Number(original.row) + 1}</>,
```
joneubank: Do we use this +1 logic in a lot of places (or even.. more than 1?) feels like there should be a clinical util that converts from submission data row to display row (so we don't have to have scattered +1's all over)

hlminh2000: yeah great point! There are exactly 2 places right now, so it makes sense to pass both through the same function. 
Will update :) only thing to note is we still need to call that function at multiple points because this should be handled only immediately at the display, anything upstream is prone to bugs.

joneubank: yes that is what i was expecting, can't modify the data as they are row indices and we want to keep them unchanged... just want to enforce the same modification at the display in all cases.

PR #959, opened 2019-12-05
✨ Ids for E2E testing of Registration

joneubank:
Small changes, adding identifiers for e2e testing to use.

### DIFF
--- a/components/pages/submission-system/join/JoinProgramLayout.tsx
+++ b/components/pages/submission-system/join/JoinProgramLayout.tsx
@@ -84,7 +84,6 @@ export default function JoinProgramLayout({
             css={css`
               margin-bottom: 30px;
             `}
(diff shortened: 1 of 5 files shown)


components/pages/submission-system/join/JoinProgramLayout.tsx
```diff
@@ -84,6 +84,7 @@ export default function JoinProgramLayout({
             css={css`
               margin-bottom: 30px;
             `}
+            onChange={() => {}}
```
hlminh2000: can avoid having to pass this by typing `onChange` as optional and doing the `()=>{}` fallback inside the `Tabs` component

joneubank: that'd be better. i'll do that

PR #969, opened 2019-12-17
💄 tag style overrides

ciaranschutte:
**Description of changes**
allow tag styling overrides

but why? 
as the uses cases of uikit are expanding the variant constants for components are too constricting

example: 
uikit tag component in doc site needs a different background.
can't just be a nested div because there is a specific border radius color
current variants of ERROR, WARNING etc don't make semantic sense even if the color was correct

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


ciaranschutte:
styled components expose full div api already.
so style={{}}... works as is

PR #1038, opened 2020-01-28
Bug/uikit select close

ciaranschutte:
**Description of changes**

HiddenSelect onBlur logic causing incorrect open state

- moves HiddenSelect onBlur logic to allow toggling of dropdown visibility state onClick
-  moves document click functionality to useEffect 
**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/form/Select/index.tsx
+++ b/uikit/form/Select/index.tsx
@@ -107,7 +107,7 @@ const Select: React.ComponentType<{
           setExpanded(false);
           onChange(e.target.value);
         }}
-        onFocus={e => {
+        onFocus={() => {
           setActive('focus');
           setExpanded(true);
         }}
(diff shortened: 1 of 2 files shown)


uikit/form/Select/index.tsx
```diff
@@ -59,12 +59,37 @@ const Select: React.ComponentType<{
...
+
+  const documentClickHandler = e => {
+    const target = e.target;
+    const wrapperNode = wrapperRef.current;
+
+    if (wrapperNode && !wrapperNode.contains(target as Node)) {
+      setExpanded(false);
+      setActive('default');
+    }
+  };
```
hlminh2000: nothing wrong with this, but I think it's a bit safer to have `documentClickHandler` scoped inside the `useEffect` callback, just to discourage calling it directly until such need becomes apparent :)

PR #1060, opened 2020-01-30
Fix/1035 error non existing program

jaserud:
**Description of changes**

- added program check when navigate to page with [shortName]
- create page now has startWithGlobalLoader option to set the global loader initially to true

**Type of Change**

- [ ] Bug
- [x] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-form/ProgramForm.tsx
+++ b/components/pages/submission-system/program-form/ProgramForm.tsx
@@ -166,7 +166,7 @@ export default function CreateProgramForm({
     }
   };
 
-  const disableButton = (seedFormData, form) => {
+  const isSameContent = (seedFormData, form) => {
     return Object.entries(seedFormData).every(([fieldName, seedVal]: [string, any[]]) => {
       const formVal = form[fieldName];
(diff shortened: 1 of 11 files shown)


components/pages/submission-system/program-form/ProgramForm.tsx
```diff
@@ -531,7 +543,7 @@ export default function CreateProgramForm({
           <Button
             id="button-submit-edit-program-form"
             onClick={submitForm}
-            disabled={isMatch(seedFormData, form)}
+            disabled={disableButton(seedFormData, form)}
```
jaserud: just checking the objects are equal wasn't enough because the order of sub-arrays could be different.

components/pages/submission-system/program-form/ProgramForm.tsx
```diff
@@ -164,6 +166,16 @@ export default function CreateProgramForm({
     }
   };
 
+  const disableButton = (seedFormData, form) => {
```
hlminh2000: I think this reads better when named `isSameContent` or something along the line, and use `!isSameContent(seedFormData, form)`

PR #1070, opened 2020-02-04
💚 Adding Button Ids on User Page

P-Saha:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/user/AccessKeyBox/index.tsx
+++ b/components/pages/user/AccessKeyBox/index.tsx
@@ -82,6 +82,7 @@ const AccessTokenBox = ({ accessKey, loading }: { accessKey: AccessKey; loading:
           `}
         >
           <div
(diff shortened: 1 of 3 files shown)


components/pages/user/AccessKeyBox/index.tsx
```diff
@@ -82,6 +82,7 @@ const AccessTokenBox = ({ accessKey, loading }: { accessKey: AccessKey; loading:
           `}
         >
           <div
+            className="accessTokenBox" // For Selenium
```
hlminh2000: classNames typically imply there can be multiple. When the intention is to address a single specific element on the page, id is usually preferred :)

PR #1077, opened 2020-02-06
Uikit/content menu

ciaranschutte:
**Description of changes**

adds content menu tracking 

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


hlminh2000:
just to capture the discussion we had :)

Let's make this component pure presentational and let the consumer manage state. That way we don't have this split logic between the doc site and uikit

ciaranschutte:
@hlminh2000 sounds good thank you Minh!

ciaranschutte:
@hlminh2000 addressed
- pulled out all stateful logic related to linking to scroll
- css specificity in here isn't harmful, bigger pain to do it in docusaurus

uikit/package.release.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@icgc-argo/uikit",
-  "version": "1.0.9",
+  "version": "1.0.10.alpha.ciaran.1",
```
joneubank: why are we adding `.ciaran` ?

ciaranschutte: @joneubank testing branch. needs to be published to npm to test because testing locally as a package is not working.

If this code is all good then we can change it to proper release eg. 1.0.10 and release that.

Might be a better way, what do you think?

ciaranschutte:
@joneubank downside of that is we have our theme colors in uikit
Something we can improve on. Using emotion + theme in `argo-docs` would make it easier.

joneubank:
> @joneubank downside of that is we have our theme colors in uikit
> Something we can improve on. Using emotion + theme in `argo-docs` would make it easier.

Ah, this is a good point. Makes me more accepting of this solution.

PR #1078, opened 2020-02-07
🐛 1067 - prevent enabled save when user modal data is not changed

anncatton:
**Description of changes**

On the EditUser modal, save button will remain disabled unless user form data is changed.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/modals/editUser/index.tsx
+++ b/components/pages/submission-system/modals/editUser/index.tsx
@@ -45,7 +45,11 @@ const EditUserModal = ({
     >
       <UserSection
         user={form}
-        onChange={(key, val) => setData({ key, val })}
+        onChange={(key, val) => {
+          if (form[key] !== val) {
+            return setData({ key, val });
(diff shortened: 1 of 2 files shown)


components/pages/submission-system/modals/editUser/index.tsx
```diff
@@ -45,7 +45,11 @@ const EditUserModal = ({
     >
       <UserSection
         user={form}
-        onChange={(key, val) => setData({ key, val })}
+        onChange={(key, val) => {
+          if (form[key] !== val) {
+            return setData({ key, val });
+          }
+        }}
```
hlminh2000: I think this logic belongs to the `useFormHook`, it computes the `touched` property of a form, which tells you if the form state differs from its initial state. my gut says the `useFormHook` isn't doing a deep diffing.

PR #1101, opened 2020-02-20
Add sleep() to loading modal to prevent flashes

wajiha-oicr:
**Description of changes**

- Whenever you would click on a program's page, while the program's existence was being checked, a loading modal was rendered
- In the case where the existence was confirmed to be true, the modal was wiped away so quickly that it looked like a strange flash on the page
- Added a 1200ms sleep before the modal removal so the loading animation is actually obvious before it gets wiped away

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/global/hooks/useProgramCheckEffect.tsx
+++ b/global/hooks/useProgramCheckEffect.tsx
@@ -1,15 +1,14 @@
 import { usePageQuery } from 'global/hooks/usePageContext';
-import { useState, useEffect } from 'react';
+import { useEffect } from 'react';
 import { useQuery } from '@apollo/react-hooks';
 import { ERROR_STATUS_KEY } from 'pages/_error';
 import PROGRAM_SHORTNAME from './PROGRAM_SHORTNAME.gql';
 import { useGlobalLoadingState } from 'components/ApplicationRoot';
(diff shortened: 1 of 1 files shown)


global/hooks/useProgramCheckEffect.tsx
```diff
@@ -1,32 +1,41 @@
...
+        throw err;
+      }
 
-  if (programNotExist) {
-    const err = new Error('Program not found') as Error & { statusCode?: number };
-    err[ERROR_STATUS_KEY] = 404;
-    throw err;
-  }
+      // Otherwise, the query result is back and the program does exist.
+      // Wait for a small buffer time so the loading modal animation has enough time.
```
hlminh2000: These comments are narrating exactly what the code is doing, which requires maintenance or it falls out of sync and becomes misleading overtime... I usually reserve comments for hidden requirements / constraints that are entirely hidden from code.

Instead, I'd suggest renaming the `isLoading` and `setLoading` to `isLoaderShown` and `setLoaderShown` as a way to make the code read better :)

PR #1121, opened 2020-02-25
Rc/1.8.17

P-Saha:


wajiha-oicr:
1.8.17 just got merged back into develop, I think you'll have to go to 1.8.18 no?

hlminh2000:
@wajiha-oicr good point! a little disappointed git doesn't catch this as a conflict...

hlminh2000:
this release will fail because 1.8.17 is already out

PR #1143, opened 2020-02-28
🐛 Remove collaborator as default new user role.

anncatton:
**Description of changes**

Removes COLLABORATOR as default role when adding a new user.
- form will display an error if a role is not selected.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/modals/common.tsx
+++ b/components/pages/submission-system/modals/common.tsx
@@ -13,7 +13,7 @@ export const UserModel = {
   firstName: '',
   lastName: '',
   email: '',
-  role: '',
+  role: null as RoleKey,
 };
 
 export const userSchema = yup.object().shape({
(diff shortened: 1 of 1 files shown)


components/pages/submission-system/modals/common.tsx
```diff
@@ -13,7 +13,7 @@ export const UserModel = {
   firstName: '',
   lastName: '',
   email: '',
-  role: 'COLLABORATOR' as RoleKey,
+  role: '',
```
hlminh2000: just wanna confirm, does `''` work with the validation logic? is `null` or `undefined` a better empty state holder? We probably want to preserve the `as RoleKey` portion for type safety so I'm thinking those would work better

anncatton: i tested the validation and it does work with the empty string - but yes, agree that `null` or `undefined` probably makes more sense.

PR #1148, opened 2020-02-28
✨Add redirect logic for public paths

anncatton:
**Description of changes**

Adds redirect logic to Navbar login button
- including for private and public paths, and no path (this should redirect to `logged-in`)
- removes `/` default value for GoogleLogin `redirectPath`, this caused a 403 on oauth because it add a trailing `/` when the original path was `/login`
- corrects a typo on the Create Program form

**Type of Change**

- [x] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/NavBar.tsx
+++ b/components/NavBar.tsx
@@ -22,6 +22,7 @@ import { getConfig } from 'global/config';
 import { createRedirectURL } from 'global/utils/common';
 import { get } from 'lodash';
 import queryString from 'query-string';
+import urlJoin from 'url-join';
 
 const NavBarLoginButton = () => {
   return (
@@ -78,7 +79,7 @@ export default function Navbar({ hideLink }: { hideLink?: boolean }) {
         path: parsedRedirect.url,
         query: existingQuery,
       });
-      setLoginPath(`${EGO_URL}${queryRedirect}`);
+      setLoginPath(urlJoin(EGO_URL, queryRedirect));
     } else if (path === '/' || path === '/login') {
       setLoginPath(EGO_URL);
     } else {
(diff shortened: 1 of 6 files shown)


components/NavBar.tsx
```diff
@@ -60,7 +63,32 @@ export default function Navbar({ hideLink }: { hideLink?: boolean }) {
   const { EGO_URL } = getConfig();
   const { token: egoJwt, logOut, data: userModel } = useAuthContext();
   const canAccessSubmission = !!egoJwt && (canReadSomeProgram(egoJwt) || isRdpcMember(egoJwt));
-  const { asPath: path } = usePageContext();
+  const { asPath: path, query } = usePageContext();
+
+  const [loginPath, setLoginPath] = React.useState('');
```
hlminh2000: I don't think this has to be a state? can just be `const loginPath = ...`. For optimization, `React.useMemo` can be used.

components/NavBar.tsx
```diff
@@ -60,7 +63,32 @@ export default function Navbar({ hideLink }: { hideLink?: boolean }) {
...
+    if (redirect) {
+      const parsedRedirect = queryString.parseUrl(redirect);
+      const existingQuery = queryString.stringify(parsedRedirect.query);
+
+      const queryRedirect = createRedirectURL({
+        origin: location.origin,
+        path: parsedRedirect.url,
+        query: existingQuery,
+      });
+      setLoginPath(`${EGO_URL}${queryRedirect}`);
```
hlminh2000: let's use `url-join` for safety :)

components/NavBar.tsx
```diff
@@ -60,7 +63,32 @@ export default function Navbar({ hideLink }: { hideLink?: boolean }) {
...
+    } else if (path === '/' || path === '/login') {
+      setLoginPath(EGO_URL);
+    } else {
+      const redirect = createRedirectURL({
+        origin: location.origin,
+        path,
+      });
+      setLoginPath(`${EGO_URL}${redirect}`);
```
hlminh2000: `url-join`

components/NavBar.tsx
anncatton: If I remove the effect I run into the issue with the location object not being defined yet...

PR #1151, opened 2020-03-02
🐛 Fix role required error display

anncatton:
**Description of changes**

Allow RoleKey to be nullable to prevent schema error when a role is not selected after opening the menu.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/global/utils/form/validations.tsx
+++ b/global/utils/form/validations.tsx
@@ -25,4 +25,5 @@ export const email = yup
 export const role = yup
   .string()
   .label('Role')
+  .nullable()
   .oneOf(PROGRAM_USER_ROLES.map(type => type.value)) as yup.StringSchema<RoleKey>;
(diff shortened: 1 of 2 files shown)


global/utils/form/validations.tsx
```diff
@@ -25,4 +25,5 @@ export const email = yup
 export const role = yup
   .string()
   .label('Role')
+  .nullable()
```
hlminh2000: is it required as well though? now that the default is removed. makes me wonder also if that has an influence on the choice between `null` and `undefined`

anncatton: Removed the `nullable()`, the field is still showing as required in the form validation even without `required()` ??

PR #1168, opened 2020-03-03
Add documentation links

P-Saha:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/global/constants/pages.tsx
+++ b/global/constants/pages.tsx
@@ -20,3 +20,11 @@ export const PROGRAM_JOIN_LOGIN_PATH = `${SUBMISSION_PATH}/program/join/login/${
 
 // rdpc path
 export const RDPC_PATH = '/rdpc';
+
+// docs paths
+export const DOCS_MANAGING_PROGRAM_ACCESS_PATH = `/docs/managing-program-access/`;
+export const DOCS_REGISTERING_SAMPLES_PATH = `/docs/registering-samples/`;
+export const DOCS_SUBMITTING_CLINICAL_DATA_PATH = `/docs/submitting-clinical-data/`;
+export const DOCS_SUBMITTED_DATA_PATH = `/docs/submitted-data/`;
+export const DOCS_SUBMITTING_MOLECULAR_DATA_PATH = `/docs/submitting-molecular-data/`;
+export const DOCS_DICTIONARY_PATH = `/dictionary/`;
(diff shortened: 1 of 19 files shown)


global/constants/pages.tsx
```diff
@@ -20,3 +20,17 @@ export const PROGRAM_JOIN_LOGIN_PATH = `${SUBMISSION_PATH}/program/join/login/${
...
+export const DOCS_REGISTERING_SAMPLES_PATH = `/docs/registering-samples/`;
+export const DOCS_SUBMITTING_CLINICAL_DATA_PATH = `/docs/submitting-clinical-data/`;
+export const DOCS_SUBMITTED_DATA_PATH = `/docs/submitted-data/`;
+export const DOCS_SUBMITTING_MOLECULAR_DATA_PATH = `/docs/submitting-molecular-data/`;
+export const DOCS_DICTIONARY_PATH = `/dictionary/`;
+export const DOCS_API_TOKEN_PATH = `/docs/user-profile-and-api-token/`;
+export const DOCS_DATA_ACCESS_PATH = `/docs/data-access/`;
+export const DOCS_SUBMISSION_OVERVIEW_PATH = `/docs/submission-overview/`;
+export const DOCS_DATA_DOWNLOAD = `/docs/data-download/`;
+
```
wajiha-oicr: maybe you could extract the /docs/ into its own 'root' variable and when declaring the other paths dependent on the /docs/, you could append it to the root?

P-Saha: This is already being done, the Docs Root is an environment variable though because it is different depending on the environment.

hlminh2000: I think @wajiha-oicr meant taking the `/docs` portion and make it something like `const DOC_PATH="/docs"` then ```const DOCS_DATA_ACCESS_PATH=`${DOC_PATH}/data-access/` ```. I agree but this it's very critical here since there isn't no use-case for just `"/docs"` right now, so not very important :P

pages/maintenance.tsx
```diff
@@ -47,7 +49,7 @@ export default createPage({
             We'll be back up and running as quickly as possible. We appreciate your patience.
           </Typography>
           <Typography variant="subtitle2">
-            If you have any questions, <Link href="/">Contact Us</Link> or head back{' '}
+            If you have any questions, <Link href={CONTACT_PAGE_PATH}>Contact Us</Link> or head back{' '}
```
hlminh2000: this is an internal link so let's wrap it with `next/link` (https://nextjs.org/docs/api-reference/next/link)

P-Saha: Fixed!

PR #1192, opened 2020-03-05
1072/same email add user

P-Saha:
**Description of changes**

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/modals/addUser/index.stories.tsx
+++ b/components/pages/submission-system/modals/addUser/index.stories.tsx
@@ -7,7 +7,7 @@ const AddUserModalStory = storiesOf(`${__dirname}`, module).add('Basic', () => (
   <AddUserModal
     onSubmit={data => action('Submit data')(data)}
     dismissModal={action('dismissModal')}
-    users={null}
+    users={[]}
   />
 ));
 
--- a/components/pages/submission-system/modals/addUser/index.stories.tsx
+++ b/components/pages/submission-system/modals/addUser/index.stories.tsx
@@ -7,6 +7,7 @@ const AddUserModalStory = storiesOf(`${__dirname}`, module).add('Basic', () => (
   <AddUserModal
     onSubmit={data => action('Submit data')(data)}
     dismissModal={action('dismissModal')}
+    users={null}
   />
 ));
 
--- a/components/pages/submission-system/modals/addUser/index.tsx
+++ b/components/pages/submission-system/modals/addUser/index.tsx
@@ -68,10 +68,12 @@ const AddUserModal = ({
   // check for errors
   const errorCheck = () => {
     const formSubKeys = Object.keys(formSubscriptions);
-    const validity = formSubKeys
+    const invalidity = formSubKeys
       .map(key => formSubscriptions[key].hasErrors)
       .reduce((acc, val) => acc || val, false);
-    setHasErrors(validity);
+    const formEmails = formSubKeys.map(key => formSubscriptions[key].data.email);
+    const emailInvalidity = new Set(formEmails).size !== formEmails.length;
+    setHasErrors(invalidity || emailInvalidity);
   };
 
   // check if last form section is touched
--- a/components/pages/submission-system/modals/addUser/index.tsx
+++ b/components/pages/submission-system/modals/addUser/index.tsx
@@ -8,16 +8,26 @@ import { UserSection, UserSectionProps } from '../styledComponents';
 import useFormHook from 'global/hooks/useFormHook';
 import { UserModel, userSchema } from '../common';
 import uniqueId from 'lodash/uniqueId';
-import isEmpty from 'lodash/isEmpty';
+import { firstName } from 'global/utils/form/validations';
 
-const AddUser = ({ id, formSubscriptions, removeSection, onUpdate, showDelete }) => {
+const AddUser = ({ id, formSubscriptions, removeSection, onUpdate, showDelete, users }) => {
   const form = useFormHook({ initialFields: UserModel, schema: userSchema });
+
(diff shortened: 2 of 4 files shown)


components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -81,16 +83,36 @@ const AddUserModal = ({
...
+    return false;
+  };
+
   // validate each individual form and fire onSubmit for each
   const submitForm = async () => {
     const allForms = Object.keys(formSubscriptions).map(async key => {
       const form = formSubscriptions[key];
       return form.validateForm(form.data);
     });
+    const allFormsDuplicateEmail = Object.keys(formSubscriptions) // True if a duplicate Email exists
```
hlminh2000: `Object.values` work better if the values are the only interest. Also `.some` reads better, so, this whole block can be reduced to:
```
Object.values(formSubscriptions)
    .some(formHasDuplicateEmail)
```

where `formHasDuplicateEmail` is `duplicateEmailValidate`, just reads better :P

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -81,16 +83,36 @@ const AddUserModal = ({
     setIsLastSectionTouched(lastSection.touched);
   };
 
+  // sets Error and returns true if an error exists
+  const duplicateEmailValidate = form => {
+    const programEmails = Object.keys(formSubscriptions)
+      .map(key => formSubscriptions[key].data.email)
+      .concat(Object.keys(users).map(key => users[key].email));
```
hlminh2000: isn't users an array? can just do `users.map(({email}) => email)`

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -81,16 +83,36 @@ const AddUserModal = ({
...
+    if (new Set(programEmails).size !== programEmails.length) {
```
hlminh2000: functionally the same, but `uniq(programEmails).length !== programEmails.length` reads better :) (https://lodash.com/docs/4.17.15#uniq)

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -68,10 +70,10 @@ const AddUserModal = ({
   // check for errors
   const errorCheck = () => {
     const formSubKeys = Object.keys(formSubscriptions);
-    const validity = formSubKeys
+    const invalidity = formSubKeys
```
hlminh2000: might as well rewrite this to `Object.values(formSubscriptions).some(form => form.hasErrors)`

components/pages/submission-system/modals/addUser/index.tsx
hlminh2000: `allFormsDuplicateEmail` doesn't sound like a boolean... `someFormHasDuplicateEmail` is more expressive :P it's already a long name, might as well reflect what it is.

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -8,9 +8,9 @@ import { UserSection, UserSectionProps } from '../styledComponents';
 import useFormHook from 'global/hooks/useFormHook';
 import { UserModel, userSchema } from '../common';
 import uniqueId from 'lodash/uniqueId';
-import isEmpty from 'lodash/isEmpty';
+import { firstName } from 'global/utils/form/validations';
 
-const AddUser = ({ id, formSubscriptions, removeSection, onUpdate, showDelete }) => {
+const AddUser = ({ id, formSubscriptions, removeSection, onUpdate, showDelete, users }) => {
```
hlminh2000: is `users` used?

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -151,6 +173,7 @@ const AddUserModal = ({
               lastSectionTouchCheck();
             }}
             showDelete={formIds.length > 1}
+            users={users}
```
hlminh2000: does `AddUser` actually uses `users`?

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -46,9 +46,11 @@ const AddSection = styled(Button)`
 const AddUserModal = ({
   onSubmit,
   dismissModal,
+  users,
 }: {
   onSubmit: (data: typeof UserModel[]) => any | void;
   dismissModal: () => any | void;
+  users: [];
```
hlminh2000: `[]` is a tuple type. This should be `Array<typeof UserModel>`

components/pages/submission-system/modals/addUser/index.stories.tsx
```diff
@@ -7,6 +7,7 @@ const AddUserModalStory = storiesOf(`${__dirname}`, module).add('Basic', () => (
   <AddUserModal
     onSubmit={data => action('Submit data')(data)}
     dismissModal={action('dismissModal')}
+    users={null}
```
hlminh2000: probably `[]` is a better empty value. or maybe give it an actual empty sample?

components/pages/submission-system/modals/addUser/index.tsx
P-Saha: When using .some, the higher order function stops at the first fail, when it is preferred to have all of the functions run for every value. I will however keep the rest in mind!

hlminh2000: gotcha. In that case, it's better to extract the side effect outside and trigger it here. Having a side effect inside a map callback is trippy...

PR #1201, opened 2020-03-10
Build Static State for Dashboard table

wajiha-oicr:
**Description of changes**
donor data summary screenshot

- Built the donor summary table static state for the dashboard page ([zepplin ref](https://app.zeplin.io/project/5a69fad2bde6330a929157fd/screen/5deeac22f471a10775935a47)) .
- Defined the columns and the cell styles
- Defined the types for the table data
- Created 5 dummy records for demonstration purposes
- Dashboard render state is set by an environment variable

What was not done / other notes:
- sorting logic for some of the complicated cells (i.e samples, alignment)
- adding the filtering buttons and submenus
- integration with any data or states
- small scope change with the samples / raw reads column. Rendered both data into one cell rather than the colspan / merge cell like the ui mockup initially had

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -1,6 +1,7 @@
 NODE_PATH=./
 NODE_ENV=development | production
 AUTH_DISABLED=false
+DASHBOARD_ENABLED=false
 
 ######### Gateway API
 GATEWAY_API_ROOT=http://localhost:9000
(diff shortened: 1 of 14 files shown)


wajiha-oicr:
> Do you think we could create a single component from `Pipe` and `Progress`?
> There seems to be a lot of overlap. Your `Pipe` works really well

You're right, they are quite similar. I used the progress component as a reference point for the pipe :P. I was actually thinking to just using the progress component, but was unsure on how I would address some of the differences between the two, i.e. the difference in the triangle separator, using text not icons as fillers, having a single item span the entire width with rounded corners.

hlminh2000:
Gonna merge so we have it ready for planning :)

PR #1202, opened 2020-03-11
🐛 Non-duplicate emails are errors if there is a different duplicate email

P-Saha:
**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/modals/addUser/index.tsx
+++ b/components/pages/submission-system/modals/addUser/index.tsx
@@ -86,12 +86,11 @@ const AddUserModal = ({
 
   // sets Error and returns true if an error exists
   const formHasDuplicateEmail = form => {
-    let programEmails = Object.keys(formSubscriptions).map(
-      key => formSubscriptions[key].data.email,
-    );
-    programEmails.splice(programEmails.findIndex(email => email === form.data.email), 1);
-    programEmails = programEmails.concat(users.map(({ email }) => email));
-    return programEmails.includes(form.data.email);
+    const formsWithThisEmail = Object.keys(formSubscriptions)
+      .map(key => formSubscriptions[key].data.email)
+      .concat(users.map(({ email }) => email))
+      .filter(email => email === form.data.email);
+    return formsWithThisEmail.length > 1;
   };
 
   // validate each individual form and fire onSubmit for each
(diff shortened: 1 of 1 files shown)


components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -86,10 +86,12 @@ const AddUserModal = ({
...
-    const programEmails = Object.keys(formSubscriptions)
-      .map(key => formSubscriptions[key].data.email)
-      .concat(users.map(({ email }) => email));
-    return uniq(programEmails).length !== programEmails.length;
+    let programEmails = Object.keys(formSubscriptions).map(
+      key => formSubscriptions[key].data.email,
+    );
+    programEmails.splice(programEmails.findIndex(email => email === form.data.email), 1);
+    programEmails = programEmails.concat(users.map(({ email }) => email));
+    return programEmails.includes(form.data.email);
```
hlminh2000: Would this work? The mutative process is a bit hard to read...
```
const programEmails = Object.values(formSubscriptions).map(form => form.data.email)
const formsWithMatchedEmail = programEmails.filter(email => email === form.data.email)
return formsWithMatchedEmail.length > 1
```

P-Saha: Will use less mutation :)

components/pages/submission-system/modals/addUser/index.tsx
```diff
@@ -86,10 +86,11 @@ const AddUserModal = ({
 
   // sets Error and returns true if an error exists
   const formHasDuplicateEmail = form => {
-    const programEmails = Object.keys(formSubscriptions)
+    const formsWithThisEmail = Object.keys(formSubscriptions)
       .map(key => formSubscriptions[key].data.email)
-      .concat(users.map(({ email }) => email));
-    return uniq(programEmails).length !== programEmails.length;
+      .concat(users.map(({ email }) => email))
```
hlminh2000: `formsWithThisEmail` is a bit misleading because we've already mapped the forms to just emails at this point. Maybe `matchingEmails` sounds better because it's really a list of emails, not forms.

PR #1220, opened 2020-03-16
Rely on default tab

wajiha-oicr:
**Description of changes**

Under a program's clinical submission, if one of the files, (i.e. `donor.tsv`) had a priority status (i.e. `ERROR`), that tab would become the `selectedClinicalEntityType`.

When jumping straight to another program's clinical submission (on its first load), where a different file has priority status (i.e. `Specimen.tsv` is set to `WARNING`) the `selectedClinicalEntityType` tab would short circuit the logic and the intended tab with priority would not be highlighted.

Removed `selectedClinicalEntityType` in the condition.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-clinical-submission/PageContent.tsx
+++ b/components/pages/submission-system/program-clinical-submission/PageContent.tsx
@@ -93,10 +93,14 @@ const gqlClinicalEntityToClinicalSubmissionEntityFile = (
 export default () => {
   const { shortName: programShortName } = usePageQuery<{ shortName: string }>();
   const { setLoading: setPageLoaderShown } = useGlobalLoadingState();
+
+  // not declared as a side effect that changes with program change
+  // change in 'data' always seems to take precedence over currentFileList within `defaultClinicalEntityType`
   const [currentFileList, setCurrentFileList] = React.useState<{
     fileList: FileList | null;
     shortName: string;
   }>({ fileList: null, shortName: programShortName });
+
   const {
     isModalShown: signOffModalShown,
     getUserConfirmation: getSignOffConfirmation,
@@ -127,6 +131,9 @@ export default () => {
       data.clinicalSubmissions.clinicalEntities
         .map(e => e.clinicalType)
         .find(entityType => !!file.name.match(new RegExp(`^${entityType}.*\\.tsv`)));
+
+    // currentfilelist state can persist when the program changes
+    // ensure currentfilelist is specific to the current program, so sorting does not get affected by different program
     const lastUploadedEntityTypes =
       currentFileList.shortName === programShortName
         ? uniq(map(currentFileList.fileList, fileToClinicalEntityType))
--- a/components/pages/submission-system/program-clinical-submission/PageContent.tsx
+++ b/components/pages/submission-system/program-clinical-submission/PageContent.tsx
@@ -165,6 +165,7 @@ export default () => {
   >(UPLOAD_CLINICAL_SUBMISSION, {
     onCompleted: () => {
       setSelectedClinicalEntityType(defaultClinicalEntityType);
+      setCurrentFileList(null);
     },
   });
   const [validateSubmission] = useMutation<
--- a/components/pages/submission-system/program-clinical-submission/PageContent.tsx
+++ b/components/pages/submission-system/program-clinical-submission/PageContent.tsx
@@ -110,7 +110,7 @@ export default () => {
     refetch,
   } = useClinicalSubmissionQuery(programShortName, {
     onCompleted: () => {
-      setSelectedClinicalEntityType(selectedClinicalEntityType || defaultClinicalEntityType);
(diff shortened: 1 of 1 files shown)


components/pages/submission-system/program-clinical-submission/PageContent.tsx
```diff
@@ -110,7 +110,7 @@ export default () => {
     refetch,
   } = useClinicalSubmissionQuery(programShortName, {
     onCompleted: () => {
-      setSelectedClinicalEntityType(selectedClinicalEntityType || defaultClinicalEntityType);
+      setSelectedClinicalEntityType(defaultClinicalEntityType);
```
hlminh2000: just wanna confirm, this does not affect the logic that happens during mutations (upload, validate, ...) right? I'm guessing `defaultClinicalEntityType` is recomputed when data changes

wajiha-oicr: Although it does get recomputed, this comment made me find another issue with the code. 

By uploading another file, `currentFileList` gets a state change. The problem is that the state seems to be shared amongst every program. So when switching back and forth between programs, when `defaultClinicalEntityType` gets recomputed, https://github.com/icgc-argo/platform-ui/blob/48a79e10c31da735cf090c50c16da385591fafaf/components/pages/submission-system/program-clinical-submission/PageContent.tsx#L127
https://github.com/icgc-argo/platform-ui/blob/48a79e10c31da735cf090c50c16da385591fafaf/components/pages/submission-system/program-clinical-submission/PageContent.tsx#L132

the `lastUploadedEntityTypes` logic overrides the proper sorting order.

wajiha-oicr: What do you think is the approach to solve this? Is it possible to make the currentFileList state unique to each program?

wajiha-oicr:
`defaultClinicalEntityType` gets recomputed, and resetting the state of `currentFileList` to null right after it's used to sort the tabs, fixes the bug.

components/pages/submission-system/program-clinical-submission/PageContent.tsx
```diff
@@ -165,6 +165,7 @@ export default () => {
   >(UPLOAD_CLINICAL_SUBMISSION, {
     onCompleted: () => {
       setSelectedClinicalEntityType(defaultClinicalEntityType);
+      setCurrentFileList(null);
```
hlminh2000: if the problem is that `currentFileList` is persisted across programs (which makes sense, since the submission page component will not be re-rendered), the solution seems to be a side effect to reset this state explicitly, observing `programShortName`. More concretely, the following:
```
React.useEffect(() => {
  setCurrentFileList(null)
}, [programShortName])
```
would this work?

calling `setCurrentFileList(null)` in the `onCompleted` callback seems a little strange, because the only other time we are calling `setCurrentFileList` is immediately before calling `uploadClinicalSubmission` (in `handleSubmissionFilesUpload`), which begs the question, why did we do that in the first place?

wajiha-oicr: I will try the sideEffect, thanks for showing me that.

I think it was called beforehand because it impacts how the tab priority sorting works after upload (so the most recently uploaded file gets priority).

wajiha-oicr: where is an appropriate time to call the useEffect, then?

hlminh2000: `useEffect` (or any react hook for that matter) should be called at the top level of the component's function definition (not inside any function, loops, if statement, etc...).

In this case, the effect is directly tied to `currentFileList`, so I would suggest immediately after `currentFileList`'s declaration.

hlminh2000: for future reference: we've also noticed `currentFileList` was not part of the dependency list for `defaultClinicalEntityType` so added it there but there was no effect

components/pages/submission-system/program-clinical-submission/PageContent.tsx
```diff
@@ -93,7 +93,14 @@ const gqlClinicalEntityToClinicalSubmissionEntityFile = (
 export default () => {
   const { shortName: programShortName } = usePageQuery<{ shortName: string }>();
   const { setLoading: setPageLoaderShown } = useGlobalLoadingState();
-  const [currentFileList, setCurrentFileList] = React.useState<FileList | null>(null);
+
+  // not declared as a side effect that changes with program change
+  // change in 'data' always seems to take precedence over currentFileList within `defaultClinicalEntityType`
```
hlminh2000: please link provide a link to this PR in the comment so we can refer back in future :)

PR #1223, opened 2020-03-17
🐛 Prevent server error where 403 should be thrown.

anncatton:
**Description of changes**

Fix server error when user tries to access program for which they do not have access.
App will throw 403 instead of 401 when isAccessible returns false
Adds 403 error page
Small styling change on error pages

**Type of Change**

- [x] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


hlminh2000:
the 403 code does make sense, but I'm not so sure how changing the error code fixes the problem, could you provide some clarification here? :)

hlminh2000:
Also we're basically handling the `401` case implicitly by forcing the user to login, so we shouldn't ever really be seeing a `401`, can just replace the error page content I think...

anncatton:
> the 403 code does make sense, but I'm not so sure how changing the error code fixes the problem, could you provide some clarification here? :)

it looks like the issue was in how we were setting the code in `_error`: we set the err statusCode, but the response statusCode needed to be changed to match the error we're throwing, or it will always be 500. We were setting `res.status` but the valid property to change was `res.statusCode`

anncatton:
> Also we're basically handling the `401` case implicitly by forcing the user to login, so we shouldn't ever really be seeing a `401`, can just replace the error page content I think...

That makes sense. I'll update it!

PR #1234, opened 2020-03-19
336 contact redesign

P-Saha:
**Description of changes**

**Type of Change**

- [ ] Bug
- [x] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -19,8 +19,7 @@ EGO_CLIENT_ID=platform
 DOCS_URL_ROOT = 'https://docs.icgc-argo.org/'
 
 ######### ReCAPTCHA
-SITE_KEY = ''
-SECRET_KEY = ''
+RECAPTCHA_SITE_KEY = ''
 
 # Testing
 BROWSERSTACK_ACCESS_KEY= ''
--- a/.env.schema
+++ b/.env.schema
@@ -18,6 +18,10 @@ EGO_CLIENT_ID=platform
 ######### Docs
 DOCS_URL_ROOT = 'https://docs.icgc-argo.org/'
 
+######### ReCAPTCHA
+SITE_KEY = ''
+SECRET_KEY = ''
+
 # Testing
 BROWSERSTACK_ACCESS_KEY= ''
 BROWSERSTACK_USER= ''
--- a/components/pages/contact.tsx
+++ b/components/pages/contact.tsx
@@ -23,11 +23,19 @@ import {
   DOCS_SUBMITTING_MOLECULAR_DATA_PATH,
   DOCS_DATA_ACCESS_PATH,
   DOCS_DATA_DOWNLOAD,
+  DACO_PATH,
 } from 'global/constants/pages';
 
 const Ul = styled('ul')`
   ${({ theme }) => css(theme.typography.paragraph)};
   padding-left: 18px;
+  margin-top: 5px;
+  margin-bottom: 30px;
+`;
+
+const Ul2 = styled('ul')`
+  ${({ theme }) => css(theme.typography.paragraph)};
+  padding-left: 0px;
   margin-top: 5px;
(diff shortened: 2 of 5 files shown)


components/pages/contact.tsx
```diff
@@ -32,63 +32,53 @@ const Ul = styled('ul')`
...
+`;
+
 const FlexRow = styled('div')`
   display: flex;
   align-items: center;
   justify-content: space-between;
   margin-bottom: 16px;
 `;
 
+function onChange(value) {
```
hlminh2000: this callback should really be encapsulated inside `ContactPage` eh, nothing else should gain access to this directly.

Also probably `onReCAPTCHAchange` to differentiate from handlers for other elements. And please provide typing for `value` :)

components/pages/contact.tsx
```diff
@@ -352,8 +352,13 @@ export default function ContactPage() {
                 </FormControl>
               </Col>
             </Row>
-            <Row>
-              <Col align="start">
+            <Row align="end">
+              <Col>
+                <ReCAPTCHA sitekey="6Lfvf-IUAAAAAGJNRhb_urfYwVVBVjOPiQQJbxuz" onChange={onChange} />
```
hlminh2000: let's pull `sitekey` from env so we can configure this for each environment

components/pages/contact.tsx
P-Saha: Fixed!

components/pages/contact.tsx
P-Saha: In the ReCAPTCHA settings you can configure multiple environments/domains at once. I have still added this as an environment variable however.

.env.schema
```diff
@@ -18,6 +18,10 @@ EGO_CLIENT_ID=platform
 ######### Docs
 DOCS_URL_ROOT = 'https://docs.icgc-argo.org/'
 
+######### ReCAPTCHA
+SITE_KEY = ''
+SECRET_KEY = ''
```
hlminh2000: `SECRET_KEY` is a secret, shouldn't be exposed to client (aka, browser). It's not used here, can we remove?

.env.schema
```diff
@@ -18,6 +18,10 @@ EGO_CLIENT_ID=platform
 ######### Docs
 DOCS_URL_ROOT = 'https://docs.icgc-argo.org/'
 
+######### ReCAPTCHA
+SITE_KEY = ''
```
hlminh2000: maybe name this `RECAPCHA_SITE_KEY` so it's clearer what it's for

.env.schema
P-Saha: 🔥

.env.schema
P-Saha: Renamed :)

global/config.ts
```diff
@@ -18,6 +18,7 @@ export const getConfig = () => {
       `/api/oauth/login/google?client_id=${publicConfig.EGO_CLIENT_ID || ''}`,
     ),
     DOCS_URL_ROOT: publicConfig.DOCS_URL_ROOT || 'https://docs.icgc-argo.org/docs/',
+    RECAPTCHA_SITE_KEY: publicConfig.SITE_KEY || '6Lfvf-IUAAAAAGJNRhb_urfYwVVBVjOPiQQJbxuz',
```
hlminh2000: oops `publicConfig.RECAPTCHA_SITE_KEY` :)

hlminh2000: also can you give this key a shot? :) `6Lebz-IUAAAAACY7eMyfK4H52Sxy9hv4FYjhFgSR` if working, let's use that as default

P-Saha: This key seems to be working, will make default!

PR #1244, opened 2020-03-20
Rc/1.9.7

hlminh2000:


P-Saha:
ReCAPTCHA doesn't seem to be working on dev, although it was working locally. The fix for this is likely on the reCAPTCHA admin side, can you look into it @hlminh2000 ?

hlminh2000:
@P-Saha there's a mispell in the config file. `publicConfig.RECATCHPHA_SITE_KEY`

PR #1245, opened 2020-03-20
🐛Clinical Error state would not show in side tab

wajiha-oicr:
**Description of changes**

- If the **first** file you try uploading in a clinical submission is an invalid one, it appears that the submission state remains as "null", ie. does not enter the `open` state

- In these cases, the side tab, then would not show an error state.

- Ensured that even in these null cases, error would show if they existed

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/SideMenu.tsx
+++ b/components/pages/submission-system/SideMenu.tsx
@@ -193,6 +193,8 @@ const LinksToProgram = (props: { program: SideMenuProgram; isCurrentlyViewed: bo
                       INVALID: <Icon name="exclamation" fill="error" width="15px" />,
                       INVALID_BY_MIGRATION: <Icon name="exclamation" fill="error" width="15px" />,
                       PENDING_APPROVAL: <Icon name="lock" fill="accent3_dark" width="15px" />,
+                      // submission state remains as null and rejects creating open state with initial invalid upload
+                      // if errors exist, error icon should still show up despite the null state
                       null: clinicalSubmissionHasSchemaErrors ? (
                         <Icon name="exclamation" fill="error" width="15px" />
                       ) : null,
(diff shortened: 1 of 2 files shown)


components/pages/submission-system/SideMenu.tsx
```diff
@@ -193,6 +193,9 @@ const LinksToProgram = (props: { program: SideMenuProgram; isCurrentlyViewed: bo
                       INVALID: <Icon name="exclamation" fill="error" width="15px" />,
                       INVALID_BY_MIGRATION: <Icon name="exclamation" fill="error" width="15px" />,
                       PENDING_APPROVAL: <Icon name="lock" fill="accent3_dark" width="15px" />,
+                      null: clinicalSubmissionHasSchemaErrors ? (
+                        <Icon name="exclamation" fill="error" width="15px" />
+                      ) : null,
```
hlminh2000: lol this makes sense only because we all know how clinical works :P maybe a small comment explaining what null means for `typeof data.clinicalSubmissions.state` would help a lot :)

hlminh2000: wait, is `"null"` a valid `typeof data.clinicalSubmissions.state`? seems a bit strange to me, doesn't it need to be `[null]` for the value `null` to be the object key? :/

wajiha-oicr: Yes, I used null because it was the value for `state` I noticed while debugging. It's a valid type for state https://github.com/icgc-argo/platform-ui/blob/fbbe718499d2d108f9e6bdfaa636e19615503cc1/components/pages/submission-system/program-clinical-submission/types.tsx#L33

hlminh2000: I meant in order for the value `null` to be used as the key in the map, it should have to be `{ [null] : ...}` , because `{ null: ... }` is equivalent to `{ "null": ...}` and `"null"` is not a valid value for `state`

PR #1278, opened 2020-03-27
✨1042 - implement refresh token

anncatton:
**Description of changes**

Adds refresh token feature to auth flow.
- will make one attempt to refresh an expired jwt, if request fails will log out
- adds check for presence of `res` in invalid jwt clause, client side requests were not deleting ego jwt properly when invalid
- (unrelated) fix in redirectUrl creation to prevent redirect to root `redirect=/`

Will need to test this *extra* thoroughly in dev, had to use a limited setup for local development.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/ApplicationRoot.tsx
+++ b/components/ApplicationRoot.tsx
@@ -21,11 +21,7 @@ import { getConfig } from 'global/config';
 import DnaLoader from 'uikit/DnaLoader';
 import GdprMessage from './GdprMessage';
 import { FadingDiv } from './Fader';
-import Cookies from 'js-cookie';
-import { EGO_JWT_KEY } from 'global/constants';
-import { setContext } from 'apollo-link-context';
-import { isValidJwt } from 'global/utils/egoJwt';
-import refreshJwt from 'global/utils/refreshJwt';
+
 /**
  * The global portal where modals will show up
  */
--- a/components/ApplicationRoot.tsx
+++ b/components/ApplicationRoot.tsx
@@ -24,6 +24,8 @@ import { FadingDiv } from './Fader';
 import Cookies from 'js-cookie';
 import { EGO_JWT_KEY } from 'global/constants';
 import { setContext } from 'apollo-link-context';
+import { isValidJwt } from 'global/utils/egoJwt';
+import refreshJwt from 'global/utils/refreshJwt';
 /**
  * The global portal where modals will show up
  */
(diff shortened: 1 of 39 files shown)


pages/_app.tsx
```diff
@@ -70,12 +70,52 @@ class Root extends App<
...
+                body: null,
+                method: 'POST',
+                mode: 'cors',
+              });
+              const newJwt = await newRefresh.text();
+              if (isValidJwt(newJwt)) {
+                Cookies.set(EGO_JWT_KEY, newJwt);
+                refreshedJwt = newJwt;
+              } else {
+                throw new Error('Unauthorized');
```
hlminh2000: should we attach the error code for Next to catch?

global/utils/common.tsx
```diff
@@ -104,6 +104,9 @@ export const createRedirectURL = ({
   path: string;
   query?: string;
 }): string => {
+  if (path === '/' || path === '/login') {
```
hlminh2000: let's use the global constant for `/login`. I think it's `LOGIN_PAGE_PATH` :)

pages/_app.tsx
```diff
@@ -70,12 +70,52 @@ class Root extends App<
...
+                credentials: 'include',
+                headers: {
+                  accept: '*/*',
+                  authorization: egoJwt || '',
+                },
+                body: null,
+                method: 'POST',
+                mode: 'cors',
+              });
+              const newJwt = await newRefresh.text();
```
hlminh2000: I find `const newJwt = await fetch(...).then(res => res.text())` to save some naming and reading efforts :P but that's just me... 

Even better if we extract the whole fetch out as a function, this whole block becomes a bit easier to read, just because it's already pretty noisy, especially since I'm guessing it'll be needed again for handling token expiry during ajax (in the Apollo client instantiation)

pages/_app.tsx
```diff
@@ -70,12 +70,52 @@ class Root extends App<
...
     if (egoJwt) {
       if (!isValidJwt(egoJwt)) {
-        removeCookie(res, EGO_JWT_KEY);
-        redirect(res, `${LOGIN_PAGE_PATH}?redirect=${encodeURI(ctx.asPath)}`);
+        if (res) {
+          removeCookie(res, EGO_JWT_KEY);
+          redirect(res, `${LOGIN_PAGE_PATH}?redirect=${encodeURI(ctx.asPath)}`);
+        } else {
+          // pass client_id to get ego api to set correct response headers
+          if (!isRefreshing) {
```
hlminh2000: in what situation will the stack reach this point with `isRefreshing` being `true`?

anncatton: Good point. This is what I had when I was first implementing this in the apollo client creation - I might still need some logic like this there to handle cases where there's many graphql request happening at once.

hlminh2000: right, definitely needs to manage parallel requests... I guess you can let them all happen, then collect the failed ones and schedule a retry after the token refresh eh?

anncatton: yeah, that's an idea! whatever we end up doing, as long as we prevent more than one `refresh` call happening, should be ok.

anncatton: @hlminh2000 With my local setup, I wasn't sure how to reproduce parallel requests accurately...I did some testing with a shorter `pollInterval` for the user profile query (300ms) and the refresh occurred correctly. Possibly it can't be fully tested until in dev?

hlminh2000: we can probably pull in a separate `useQuery` in the same page, it should make a separate request in parallel. Can probably mess with cache policy and/or enable polling to make sure it makes the request rather than going to cache  :)

components/ApplicationRoot.tsx
```diff
@@ -147,17 +150,23 @@ const ApolloClientProvider: React.ComponentType<{ egoJwt: string; apolloCache: a
   apolloCache,
 }) => {
   const { GATEWAY_API_ROOT } = getConfig();
+  const { fetchWithRefresh } = useAuthContext();
+
+  const headerLink = setContext((request, previousContext) => ({
+    headers: {
+      ...previousContext.headers,
+      authorization: `Bearer ${Cookies.get(EGO_JWT_KEY) || ''}`,
```
hlminh2000, line 158: can we use the token coming from `useAuthContext` to let `useAuthContext` as single source of truth on the jwt?

global/hooks/useAuthContext.tsx
```diff
@@ -48,14 +51,63 @@ export function AuthProvider({
...
+              options.headers.authorization = `Bearer ${newJwt || ''}`;
+            } else {
+              logOut();
+            }
+
+            return fetch(uri, options);
+          });
+        }
+
+        var result: any = {};
```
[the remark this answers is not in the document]
anncatton: lol i missed that one!

global/utils/refreshJwt.ts
```diff
@@ -0,0 +1,21 @@
...
+export default egoJwt =>
+  fetch(refreshUrl, {
+    credentials: 'include',
+    headers: {
+      accept: '*/*',
+      authorization: egoJwt || '',
+    },
+    body: null,
+    method: 'POST',
+    mode: 'cors',
```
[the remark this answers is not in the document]
anncatton: According to mdn, the body is still readable, it just limits the response headers: https://developer.mozilla.org/en-US/docs/Web/API/Request/mode
Out of curiosity, i commented that line out, and the request headers still show `Sec-Fetch-Mode: cors`

components/ApplicationRoot.tsx
anncatton: I'm not sure why, but using the token here is not getting the updated token, even though logging in App Root shows the token is updated to match the current cookie after refresh. I'm guessing some kind of timing issue? Because it looks as if `setContext` is working as expected.

PR #1290, opened 2020-04-01
✨ Beep Boop adding robots.txt

P-Saha:
**Description of changes**
Adding robots.txt

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,0 +1,7 @@
+# robots.txt file for ICGC ARGO
+
+User-agent: *
+Disallow: /submission
+Disallow: /logged-in
+Disallow: /user
+Disallow: /contact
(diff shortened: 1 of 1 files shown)


public/robots.txt
```diff
@@ -0,0 +1,7 @@
+# robots.txt file for ICGC ARGO
+
+User-agent: *
+Disallow: /submission
+Disallow: /logged-in
+Disallow: /user
+Disallow: /contact
```
hlminh2000, line 7: whitelisting is probably easier than blacklisting for us in the future, as most things for us are disallowed... We'd rather miss pages that should be allowed than missing pages that should be disallowed, because the cost of missing the disallow is potentially much more expensive :P

hlminh2000: in other words, just disallow everything, then we can add `Allow` directives as we go for pages we care for google to index ;)

P-Saha: Sounds good!

PR #1305, opened 2020-04-07
uikit component for file repository table

wajiha-oicr:
**Description of changes**

- Made uikit component for file repository table, which is a select table
- Also added its parent container and download buttons
- Wrote Story with random generated data

An Issue...

- ~Selecting files are triggering the table's internal state change ( not entirely sure why, I don't think this is the intended behaviour)~
- ~These changes force the table to re-render to its original format
   -meaning, user changes such as changing the page, applying a filter, resizing a column, get reset to default after a new selection is made~

- fixed by https://github.com/icgc-argo/platform-ui/pull/1307 (thanks Minh)

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


wajiha-oicr:
I attempted to circumvent the issue but was unable to get a working solution. https://react-table.js.org/faq mentions brings up a similar problem but it's V7, not V6 of react-table. 

Considering that the legitimate data being fed into the table may be fed in differently than how the other tables are done, my only guess is to maintain the state of data and within the user change callbacks do a deep comparison of the states and reject re-renders for change in the 'selectedIds' prop. I think this would also require logic on what data changes warrant a legitimate rerender.

This is a strange bug, I had attempted managing the table's render state manually to stop rerenders but wasn't able to get far.

hlminh2000:
CI is failing to build uikit... the `FileRepositoryTable` component is specific to a page so it's not really a uikit component, let's move it somewhere in the `components` directory :)

wajiha-oicr:
> CI is failing to build uikit... the `FileRepositoryTable` component is specific to a page so it's not really a uikit component, let's move it somewhere in the `components` directory :)

thanks Minh, will do

uikit/FileRepositoryTable/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+  instructionBoxButtonContentStyle,
+  instructionBoxButtonIconStyle,
+} from '../../components/pages/submission-system/common';
+import Container from 'uikit/Container';
+import Typography from 'uikit/Typography';
+import Icon from 'uikit/Icon';
+
+export default ({
+  fileRepoEntries,
+  userHasAccess,
```
hlminh2000: The user access will actually be on a per-file level so it should really be a field in the file object actually :P 

Users may have access to download some files but not others ;)

wajiha-oicr: ah okay I thought the only case was that you had be logged-in :P 
I suppose there's different tooltip text for different cases (logged in vs no access). will make the permission to download a file level property then. thanks 👍

PR #1316, opened 2020-04-08
1271 donor summary card

P-Saha:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


components/pages/submission-system/donor-dashboard/DonorSummaryCard/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+export default () => {
+  return (
+    <div>
+      <Container
+        css={css`
+          padding: 10px;
+        `}
+      >
+        <Row>
+          <Col xl={2.4} lg={4} md={12}>
```
hlminh2000, line 204: Are non-integers valid values? This is a tricky layout so I'd consult with Kim about how we want the content to flow in smaller screen, because it probably won't lay out as nicely as it looks in the design.

Maybe check out how the program dashboard is done and see if it makes sense to do something similar. For the last two elements, relative positioning is important there so the 12-column system might not apply there... the design isn't written in stone so work with Kim to find alternatives if needed :)

P-Saha: @hlminh2000 
Non-integers are valid! Have been testing it out :p 
Have been talking with Kim a bit about this ticket and have gotten a bit of feedback from her except in terms of how it should look when resized in terms of responsiveness. What do you mean by relative positioning for the last two elements?

hlminh2000: `Variant Calling` has to go on the right of `Alignment` or the arrow doesn't make sense :P

P-Saha: This is always accounted for :)

hlminh2000:
can we add a story for the component for preview please? :)

components/pages/submission-system/donor/DonorSummaryStats/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+      </Row>
+      <Row nogutter>{children}</Row>
+    </StatDesc>
+  </div>
+);
+
+const Pipeline = (stats: Record<PIPELINE_STATUS, number>) => {
+  const theme = useTheme();
+
+  const getBackgroundColour = (state: keyof Record<PIPELINE_STATUS, number>) => {
```
wajiha-oicr, line 65: this looks similar to the function used in the donor summary table. Maybe we good extract the common parts in a common.tsx?

P-Saha: Sounds good! Thanks :)

PR #1322, opened 2020-04-09
✨1266 file repo barcharts

anncatton:
**Description of changes**

Creates component for file repo charts in storybook, will be moved into platform ui code when File Repository is created
- using mock data for the time being
- updates Tooltip with new arrow design
- adds file icon

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <Bar
+              style={{
+                width: 1,
+                height: maxBy(data, 'count').count,
+                backgroundColor: theme.colors.grey_2,
+              }}
+            />
+            {orderBy(data, 'count', 'desc').map(({ category, count }) => (
+              <Tooltip
+                key={`bar-${category}`}
```
ciaranschutte, line 189: I've never seen an actual key other than `key={index}`  👍

uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+`;
+
+const FlexColumn = styled('div')`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+`;
+
+const AxisText = styled(Typography)`
+  textalign: right;
```
ciaranschutte, line 72: textalign => `text-align`

uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  return (
+    <FlexColumn
+      css={css`
+        font-size: 10px;
+        color: 1px solid ${theme.colors.grey_2};
+        height: ${chartHeight}px;
+      `}
+    >
+      <AxisText color={theme.colors.grey}>{max}</AxisText>
+      <br />
```
ciaranschutte, line 87: Not sure the <br/> is needed. Spacing covered by parent flex cont?

uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  font-size: 10px;
+  margin: 0;
+`;
+
+const YAxis = ({ max = 600, theme }) => {
+  return (
+    <FlexColumn
+      css={css`
+        font-size: 10px;
+        color: 1px solid ${theme.colors.grey_2};
```
ciaranschutte, line 82: Think we can remove this, looks like previous border code?

uikit/SimpleBarChart/index.tsx
anncatton: 🤦

uikit/SimpleBarChart/index.tsx
anncatton: the bottom margin is aligning the lower axis number with the bars - padding could work too. Not sure if there's another solution?

uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+          </FlexRow>
+        </div>
+        <FlexRow>
+          <Typography
+            css={css`
+              font-size: 12px;
+              margin: 0;
+            `}
+          >
+            1.66 TB
```
hlminh2000, line 160: this needs to be dynamic eh?

uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+const getBarHeight = (value: number): number => {
+  return (value / chartHeight) * 100;
+};
+
+const getBarWidth = (data: Array<any>): number => {
+  return Math.floor(chartWidth / data.length) - 2;
+};
+
+const chartTypeMeta = {
```
hlminh2000, line 34: just to prevent mistakes :) maybe type this guy as something like
```
{ [k: FileRepoDataType] : { ... } }
```

uikit/Tooltip/index.tsx
```diff
@@ -14,7 +14,7 @@ type TooltipProps = {
   open?: boolean;
   useContext?: boolean;
   onRequestClose?: (e: any) => any;
-  position?: 'top' | 'bottom' | 'left' | 'right';
+  position: 'top' | 'bottom' | 'left' | 'right';
```
hlminh2000, line 17: hmmm this is technically a breaking change eh? is it necessary to change the type? also a little counter-intuitive to have a required property that also has a default eh

uikit/Tooltip/index.tsx
```diff
@@ -64,7 +109,25 @@ const Tooltip: React.ComponentType<TooltipProps> = ({ html, ...rest }) => {
...
+              enabled: false,
+            },
+            flip: {
+              enabled: false,
+            },
+            hide: {
+              enabled: false,
+            },
+          },
+          ...rest.popperOptions,
```
hlminh2000, line 125: might be better to use `_.merge` for a deep merge

uikit/SimpleBarChart/index.tsx
anncatton: Yup! Will mock that

uikit/Tooltip/index.tsx
anncatton: That's true. I guess there are 2 options: 
1) still allow a tooltip without an arrow, and leave the position as optional. And it wouldn't show up unless position is specified (For the platform rosi wanted all the tooltips changed to the new style.)
2) I can get rid of the default and make it required.
 What do you think?

hlminh2000: does `position` also set the tooltip position or just the arrow? if the two are in sync, I think it makes sense to keep it optional and provide a default (that'll guarantee everything has it). 

If they're not, then it makes sense to make it required (because it's the app's responsibility to keep them in sync)

hlminh2000: in the later case though, maybe `arrowPosition` is a better prop name :)

anncatton: It does set the tooltip position as well, so they'd be in sync :)

uikit/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  program: {
+    title: 'Programs',
+    color: '#00b3d3',
```
hlminh2000, line 37: let's pull these colors from theme :) might need to bring this object into the component to access theme from `useTheme` though

uikit/SimpleBarChart/index.tsx
anncatton: Disregard...I actually got rid of that styling

components/pages/file-repository/SimpleBarChart/stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  // calculations for data count and size is tbd, will depend on chart type
+  const totalCount = data.dataTypes[currentDisplay].reduce(
+    (acc, { category, count }) => acc + count,
+    0,
+  );
+  return (
+    <SimpleBarChart
+      data={data.dataTypes[currentDisplay]}
+      type={currentDisplay}
+      totalDataSize={currentDisplay !== 'program' && totalCount}
```
hlminh2000, line 32: lol I guess typescript was never told to type check storybook, but technically `totalCount` is a number so will fail :P Anyway, not a big deal :)

components/pages/file-repository/SimpleBarChart/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+      `}
+    >
+      <AxisText color={theme.colors.grey}>{max}</AxisText>
+      <br />
+      <AxisText color={theme.colors.grey}>0</AxisText>
+    </FlexColumn>
+  );
+};
+
+const SimpleBarChart: React.ComponentType<SimpleBarChartProps> = ({
```
hlminh2000, line 103: I almost want to name this something like "CardWithBarChart" :P

anncatton: Yeah I'm not convinced it's the best name, it's not even a bar chart per se...it's like a SparseBarChart :)

components/pages/file-repository/SimpleBarChart/stories.tsx
anncatton: Oh whoops that's true...i guess once i implement this for reals i can correct this :)

PR #1331, opened 2020-04-15
Build facet and integer range component

wajiha-oicr:
**Description of changes**

- Created UIKit Component for the Facet
    - Made it extend upon the submenu component 
        - Added optional param where a clickable search icon could be added, which would then toggle a searchbar to showup
       - Added optional param where the search query state could be fed in ( needed to be live in the same place as the logic for the facet options)
        - optional param to adjust position of the chevron 
    - Created Facet Option display behaviour
        - at most, 5 options render by default, any (if any )remaining are hidden
        - more / less could be toggled ( the toggle is visible only if applicable )
        - selected options get sorted and 'pinned' to the top, no longer allowed to be hidden
        - ability to select all / deselect all
        - search bar filters options ( logic shamelessly taken from programs search)

- Created UIKit Component for the Integer Range 
    - Made it extend upon the submenu component 
    - built with Input element
    - entries are restricted to digits or empty string (backspace to empty state)
    - add logic to the button to only trigger if at least one input field has a valid value

- New Icon , new variant to tag

<sub> things I learned the hard way... 
    - declaring a styled component in the same scope where it's used and then putting an input element inside it breaks things. The input field loses focus after every keystroke https://github.com/styled-components/styled-components/issues/540
    - using type="number" on an input field means minus ( - ) signs get registered as empty strings (?????). This makes it impossible to validate for the difference between someone backspacing the entire entry and someone typing "55525--24-24-2" https://github.com/redux-form/redux-form/issues/2839#issuecomment-296524572
</sub>
**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


uikit/Facet/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+          <div
+            css={css`
+              border-top: 1px solid;
+              border-color: ${useTheme().colors.grey_2};
+            `}
+          >
+            {visibleOptions}
+          </div>
+        }
+        {viewAmountController}
```
hlminh2000, line 194: this portion of the component will be very useful on its own outside of this context, possible to split it out to a different component?

uikit/Facet/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+      ({ name }) => !searchQueryState.length || name.search(new RegExp(searchQueryState, 'i')) > -1,
+    )
+    .map(option => option.name);
+
+  const [selectedFacetOptions, setSelectedFacetOptions] = React.useState([]);
+  const [selectAll, setSelectAll] = React.useState(true);
+
+  const defaultRenderLimit = 5;
+
+  const buildStyledOption = (option: FilterOption) => {
```
hlminh2000, line 32: it's more idiomatic to write this as a component and use it with a jsx tag :)

uikit/Facet/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+          onClick={e => {
+            setAllOptionsVisible(!allOptionsVisible);
+          }}
+        >
+          <Icon
+            name={allOptionsVisible ? 'minus_circle' : 'plus_circle'}
+            css={css`
+              margin-right: 6px;
+            `}
+            fill={useTheme().colors.accent2}
```
hlminh2000, line 165: It's working fine in this case, but not recommended to write hooks inline like so, just because it's a bit hard to tell if the [rule of hooks](https://reactjs.org/docs/hooks-rules.html) are satisfied. If a conditional render is added, for a part of the tree that contains this node, the rule would be violated.

I recommend `const theme = useTheme()` up top :)

btw, by convention, anything named `use...` is a react hook ;)

uikit/IntegerFieldDisplay/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const { min, max } = getRangeValues();
+
+    if (min > max) {
+      console.log('Invalid Range. Max must be larger than Min.');
+    } else {
+      console.log(`Running search between ${min} and ${max}`);
+    }
+  };
+
+  const numberSelector = (
```
hlminh2000, line 53: likewise with the facets, this part is a very useful thing to have as a component on its own :) there will be other parts of the app where we need this setup

uikit/Facet/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+import Typography from 'uikit/Typography';
+import useTheme from 'uikit/utils/useTheme';
+import orderBy from 'lodash/orderBy';
+
+import HyperLink from 'uikit/Link';
+
+export type FilterOption = {
+  name: string;
+  quantity: number;
+};
```
hlminh2000, line 16: the data coming from Elasticsearch will probably look like the following, so if we match this now we'll save a lot of run-time loops later :) :
```
{
    key: string,
    doc_count: number
}
```

uikit/Facet/index.tsx
wajiha-oicr: the reason I initially didn't do it that was was because this function returns an object not just containing the jsx element but also two other properties ` name` and `isOptionChecked` that are used to help determine whether the jsx element is visible or not. I needed to be able to manage the components.

uikit/Facet/index.tsx
wajiha-oicr: Sure. Do you think this and the Integer range selector are unique enough that they deserve to live in their own UIKit component? Or just as exported components in a common.tsx file type of thing?

hlminh2000: yeah these will make great uikit components :) we will have uses for them very soon outside of the file repo page :)

uikit/Facet/index.tsx
hlminh2000: if we just had a top-level state of type `Array<{isChecked: boolean, name: string}>` called say `optionStates`, then we can do the following just inline and :
```
orderBy(optionStates, ['isChecked', 'name'])
  .filter(({isChecked}) => isChecked)
  .map(({name, isChecked})  => <StyledOption key={name} name={name} checked={isChecked} />)
```
where `StyledOption` is the rendering component. That way we keep the model and the view separate :)

the setter will look like 

```
const selectOption = (optionName) => setOptionState(optionStates.map((state) => ({
  ...state,
  isChecked: state.name === optionName ? !state.isChecked : state.isChecked
}))
```

wajiha-oicr:
> It's looking great! Here are some code organization stuff to make this more reusable in the short future :)

thanks for your review :) requested changes implemented here https://github.com/icgc-argo/platform-ui/commit/881ecdb7545979f58fcd38bdbb703d904d901ea1

uikit/Facet/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    )
+    .map(option => option.key);
+
+  const [selectAll, setSelectAll] = React.useState(true);
+
+  const initalStates = options.map(option => {
+    return { ...option, isChecked: false };
+  });
+
+  const [optionStates, setOptionStates] = React.useState<Array<SelectableFilterOption>>(
```
hlminh2000, line 36: states normally are meant to be serializable data, so let's avoid this where possible :) 

In this case, the `useEffect` can be replaced with directly rendering the `SelectableFilterOption` array, like so:
```
const sortedOptions = orderBy(
  optionStates,
  ["isChecked", "key"],
  ["desc", "asc"]
);

const selectedOptions = optionStates.filter((option) => option.isChecked);

const queriedOptions = sortedOptions.filter(
  option => queriedOptionKeys.includes(option.key) && !selectedOptions.includes(option),
);

const defaultNonSelectedOptions = sortedOptions
  .slice(0, defaultRenderLimit)
  .filter(option => !option.isChecked);

const optionsToShow = !!searchQueryState
  ? selectedOptions.concat(queriedOptions)
  : allOptionsVisible
  ? sortedOptions
  : selectedOptions.concat(defaultNonSelectedOptions);

return (
  <...>
    {optionsToShow.map(option => <SelectableFilterOption .../>)}
  </...>
)
```
as an optimization, each of the step of the computation can be memoized using [React.useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) with the appropriate dependency so that they're not all being run all the time :)

This is really just to stay idiomatic to the framework, what you have totally works ;)

uikit/Facet/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+            setAllOptionsVisible(!allOptionsVisible);
+          }}
+          selectAllState={selectAll}
+          toggleVisiblityCss={
+            numberofMoreOptions === 0 ? (allOptionsVisible ? 'visible' : 'hidden') : 'visible'
+          }
+          toggleText={allOptionsVisible ? `Less` : `${numberofMoreOptions} More`}
+          moreOptionsAvailable={!allOptionsVisible}
+        />
+      </MenuItem>
```
hlminh2000, line 169: sorry if I wasn't very clear in my previous comment, it's actually this bundle of the component that would be great to be packaged together into its own thing :) Here's a sample of the other usage:

Screen Shot 2020-02-05 at 12 07 28 PM

It doesn't look exactly the same but we can modify the component later to fit, but having it as a separate component now will save some refactoring work later :P

uikit/NumberRangeField/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { FieldDescriptionLabel, FieldInputWrapper } from './common';
+
+import Input from 'uikit/form/Input';
+import Button from 'uikit/Button';
+
+const inputHandler = (
+  inputVal: string,
+  stateSetter: React.Dispatch<React.SetStateAction<string>>,
+) => {
+  const validator = new RegExp(/(^$|^\d+$)/);
```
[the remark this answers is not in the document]
wajiha-oicr: yeah I went down that rabbithole, there's a bug when you set input type to number. 

so when you set type to number, you can set a min/max property. the validation isn't very robust thought and it doesn't prevent the user typing (or pasting) a negative number into the field.

But if you try to provide additional validation for negative numbers (i.e. don't allow minus signs), the odd thing is that when typing a negative sign, the input registered always defaults to an empty string. empty strings need to be accepted because otherwise, the user can't backspace, or the field can't be empty by default. This means using type="number" lets you enter something like 41234324--234-42-34-234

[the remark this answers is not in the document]
hlminh2000: I see, yeah there'll need to be more error handling in that case eh.

For the validation though, should use `isNan`? Cause things like `3.125e7` are numbers too. And negative numbers should be allowed too eh? Only special case is `""` cause `isNan("")` is `false` lol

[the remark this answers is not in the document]
wajiha-oicr: I didn't want to include negative numbers in the context that I was working with but I suppose you're right, there may be cases where those are allowed. The component was integer range field but in the future yes exponentials, decimals, negatives, etc could be allowed in different usages. 

Should I drop the specific validation? Make it a prop so each use case can be fine tuned? Or allow any input but immediately validate after the user loses focus out of the field, like how Yup is doing on the "Number of Donors" field for the manage program?

uikit/OptionsList/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  searchQuery?: string;
+  defaultRenderLimit?: number;
+}> = ({ options, searchQuery = '', defaultRenderLimit = 5 }) => {
+  const theme = useTheme();
+  const [allOptionsVisible, setAllOptionsVisible] = React.useState(false);
+
+  const queriedOptionKeys = options
+    .filter(({ key }) => !searchQuery.length || key.search(new RegExp(searchQuery, 'i')) > -1)
+    .map(option => option.key);
+
```
hlminh2000, line 31: I think memoizing this will be helpful for handling larger quantities of entries :)

Somewhere under the hood, the regex check potentially needs to loop through every character of the `key` so 200 entries with 10 characters each can end up being 2000 passes of some sort :P

PR #1341, opened 2020-04-20
1270 donor file card

P-Saha:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/donor/DonorFileCard/index.tsx
+++ b/components/pages/submission-system/donor/DonorFileCard/index.tsx
@@ -123,7 +123,9 @@ export default () => {
     >
       <Row justify="between">
         <Col>
-          <Typography variant="subtitle">Available Files</Typography>
+          <Typography variant="subtitle" as="h2">
+            Available Files
+          </Typography>
         </Col>
         <Col
           css={css`
--- a/components/pages/submission-system/donor/DonorFileCard/index.tsx
+++ b/components/pages/submission-system/donor/DonorFileCard/index.tsx
@@ -21,6 +21,7 @@ const DataRow: React.ComponentType<{ name: string; link: string; fileCount: numb
     `}
   >
     <Col
+      xs={10}
       css={css`
         white-space: nowrap;
         align-self: center;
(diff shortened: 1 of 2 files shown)


components/pages/submission-system/donor/DonorFileCard/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      variant="STATIC"
+      parentRef={{ current: null }}
+      showPagination={false}
+      data={data}
+      columns={[
+        {
+          sortable: false,
+          Header: header.toUpperCase(),
+          headerStyle: {
+            background: useTheme().colors.secondary_4,
```
hlminh2000, line 103: just a code style thing, using hooks inline in jsx is a little error prone because it can be hard to tell if the hook is following the rules of hooks. I recommend doing `const theme = useTheme()` and then use `theme` in the jsx :)

components/pages/submission-system/donor/DonorFileCard/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+          style: { whiteSpace: 'unset' },
+        },
+      ]}
+    />
+  </Col>
+);
+
+export default () => {
+  return (
+    <div>
```
hlminh2000, line 115: no need for this div eh?

components/pages/submission-system/donor/DonorFileCard/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      <Container
+        css={css`
+          padding: 20px;
+          padding-top: 0px;
+          min-width: 370px;
```
hlminh2000, line 120: this `min-width` might not play well with the other components in the page if we actually need this component to get even smaller, so I wouldn't have it here for now :)

components/pages/submission-system/donor/DonorFileCard/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+                  padding-left: 3px;
+                  position: relative;
+                  top: 1px;
+                `}
+                name="chevron_right"
+                fill="accent2_dark"
+                height="12px"
+              />
+            </Button>
+          </Col>
```
hlminh2000, line 185: I think it makes more sense to just have 1 column containing both buttons in a row eh? that way we have both buttons just moved all the way to the end.

PR #1343, opened 2020-04-20
Rc/1.8.0

hlminh2000:


jaserud:
Does that make sense??? https://github.com/icgc-argo/platform-ui/blob/f2f8a9e9e9b8c20dc4689829e5a46cd1c14735f8/package.json#L3

hlminh2000:
> Does that make sense???
> 
> https://github.com/icgc-argo/platform-ui/blob/f2f8a9e9e9b8c20dc4689829e5a46cd1c14735f8/package.json#L3

lol yeah I like the Bashar approach to version tagging :P but what do you think?

PR #1344, opened 2020-04-20
🐛 Prevent undefined redirect path

anncatton:
**Description of changes**

Prevents creating redirect when ctx path is undefined.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/global/utils/common.tsx
+++ b/global/utils/common.tsx
@@ -105,7 +105,7 @@ export const createRedirectURL = ({
   path: string;
   query?: string;
 }): string => {
-  if (path === '/' || path === LOGIN_PAGE_PATH || path === 'undefined') {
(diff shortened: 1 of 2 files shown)


global/utils/common.tsx
```diff
@@ -105,7 +105,7 @@ export const createRedirectURL = ({
   path: string;
   query?: string;
 }): string => {
-  if (path === '/' || path === LOGIN_PAGE_PATH) {
+  if (path === '/' || path === LOGIN_PAGE_PATH || path === 'undefined') {
```
hlminh2000, line 108: How would `[ "/", LOGIN_PAGE_PATH, "undefined" ].includes(path)` read for you? :)

anncatton: that looks better to me, will change :)

PR #1362, opened 2020-04-22
Feat/clinical timeline mock

ciaranschutte:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


components/pages/submission-system/donor/ClinicalTimeline/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+  { type: 'specimen', id: 'SPECIMEN SP0032', description: 'Tumour' },
+  { type: 'specimen', id: 'SPECIMEN SP2123', description: 'Tumour' },
+  { type: 'specimen', id: 'SPECIMEN SP0123', description: 'Tumour' },
+  { type: 'treatment', id: 'TREATMENT TR8982', description: 'Chemotherapy' },
+  { type: 'treatment', id: 'TREATMENT TR8982', description: 'Ablation' },
+  { type: 'treatment', id: 'TREATMENT TR8982', description: 'Loco-regional progression' },
+  { type: 'follow_up', id: 'FOLLOW UP FO2123', description: 'Relapse' },
+];
+
+const Header = ({ entityCounts, activeEntities, setFilters }) => {
```
hlminh2000, line 52: let's add some typings :)

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
+import Typography from 'uikit/Typography';
+import Container from 'uikit/Container';
+import Checkbox from 'uikit/form/Checkbox';
+import { css } from 'uikit';
+import useTheme from 'uikit/utils/useTheme';
+import colors from 'uikit/theme/defaultTheme/colors';
+
+const TIMELINE_TYPES = Object.freeze({
```
hlminh2000, line 9: I think we should avoid using things from defaultTheme directly because it removes the ability to dynamically switch theme (in case that becomes a need)... types are fine because `defaultTheme` gives the basic shape that all themes must follow anyway...

for this case I think we can update this to `(theme: typeof defaultTheme) => ...` and call it in the component :)

ciaranschutte: @hlminh2000  added a function to return the colors dynamically because colors will be used in Timeline vertical tabs component too

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+  entityCounts: { [k in EntityType]: number };
+  activeEntities: Array<EntityType>;
+  setFilters: (e: Array<EntityType>) => void;
+};
+
+const Header = ({ entityCounts, activeEntities, setFilters }: HeaderTypes) => {
+  const theme = useTheme();
+  const counts: Array<EntityType> = Object.keys(entityCounts) as Array<EntityType>;
+
+  const timelineStyles = React.useMemo(() => getTimelineStyles(theme), [theme]);
```
hlminh2000, line 78: 👍 👍 👍 👍

PR #1363, opened 2020-04-22
✨ Init file repo page with feature flag

anncatton:
**Description of changes**

Creates file repository page with route `/repository`
Adds some responsiveness to bar charts
*ui only - no functionality for facet/table search or filtering*

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


components/pages/file-repository/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+        `}
+        // {...inputProps}
+      />
+      <Icon name="upload" height="12px" fill={disabled ? 'white' : 'accent2_dark'} />
+      {' Upload a list of ids'}
+    </Button>
+  );
+};
+
+const FacetPanel = ({ theme, facets }) => {
```
hlminh2000, line 85: can just pull theme from `useTheme`, why provide as props? let's keep the props clean for meaningful things :)

components/pages/file-repository/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+  flex-direction: column;
+  justify-content: space-between;
+
+  height: calc(100vh - 58px);
+  max-height: calc(100vh - 58px);
+  overflow-y: auto;
+`;
+
+// TODO: implement search and upload functionality
+const UploadIdsButton = () => {
```
hlminh2000, line 57: there's actually a `FileSelectButton` component :) https://argo-ui-storybook.netlify.app/?path=/story/uikit-fileselectbutton--basic

components/pages/file-repository/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+          chevronOnLeftSide
+        >
+          <Input
+            css={css`
+              margin: 8px;
+            `}
+            icon={<Icon name="search" />}
+            aria-label={'Search'}
+            placeholder="e.g. FL9998, DO9898..."
+          />
```
hlminh2000, line 114: there's a `"search"` preset for the intput to ensure style and implementation consistency :) checkout the knobs https://argo-ui-storybook.netlify.app/?path=/story/uikit-form-input--with-preset

components/pages/file-repository/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+            facets={[
+              'program',
+              'primary site',
+              'age at diagnosis',
+              'vital status',
+              'gender',
+              'experimental strategy',
+              'data type',
+            ]}
+          />
```
hlminh2000, line 165: I'm not sure this list will dynamically change much, at the same time, the `FacetPanel` component will likely grow pretty big with logic, should we just split it to a separate file and bundle this list in to prevent this file from growing in the future? 

probably will have its own graphql query too so probably makes sense in its own folder...

components/pages/file-repository/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+              <PaddedColumn xl={4} lg={6} md={12}>
+                <SimpleBarChart
+                  data={dataTypes['data type']}
+                  type={'data type'}
+                  totalDataSize={`${dataTypes['data type'].reduce(
+                    (acc, item) => item.count + acc,
+                    0,
+                  )}`}
+                  totalCount={dataTypes['data type'].length}
+                />
```
hlminh2000, line 204: each of these chart will probably have its own graphql query and data transformation logic, which I think make sense to be in individual bundles. Should we have a small folder to wrap each chart?

components/pages/file-repository/index.tsx
anncatton: ok cool - i wasn't sure if that was the same functionality so i just put in a placeholder button for now. I can add it :)

components/pages/file-repository/index.tsx
anncatton: yeah for now i was just putting it in here, since there is no functionality yet. I was thinking those files would be created once we start adding in the data hookup. But I can create the structure now - what do you think?

hlminh2000: I think generally having the structure defined early will let us work better in parallel later, so many people can work on the same page without causing conflicts :)

pages/repository/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+import FileRepositoryPage from 'components/pages/file-repository';
+import { dummyData } from 'components/pages/file-repository/FileTable/dummyData';
+import useAuthContext from 'global/hooks/useAuthContext';
+
+export default createPage<{ egoJwt: string }>({
+  isPublic: true,
+  getInitialProps: async ({ egoJwt }) => ({ egoJwt }),
+})(({ egoJwt }) => {
+  const { token } = useAuthContext();
+  return <FileRepositoryPage token={token} data={dummyData} />;
```
hlminh2000, line 11: it'll probably be up to `FileRepositoryPage` to pull its own data so no need to provide data. also token should be accessed with the auth context anyway so I don't think that's needed either

components/pages/file-repository/index.tsx
anncatton: sounds good! will do

PR #1377, opened 2020-04-24
🐛 Prevent NaN fill amount on percentBar

anncatton:
**Description of changes**

Percent bar set to 0% fill amount when denominator is `0`, prevent NaN result.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/PercentBar/index.tsx
+++ b/uikit/PercentBar/index.tsx
@@ -10,8 +10,9 @@ const PercentBar: React.ComponentType<{
   const theme = useTheme();
 
   // Negative Numbers should be zero, percentages over 100 should be capped
+  // A zero denominator should show as 0% fill
   num < 0 ? (num = 0) : { num };
-  const fraction = Math.min((num / den) * 100, 100);
+  const fraction = den <= 0 ? 0 : Math.min((num / den) * 100, 100);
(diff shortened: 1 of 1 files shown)


uikit/PercentBar/index.tsx
```diff
@@ -10,8 +10,9 @@ const PercentBar: React.ComponentType<{
   const theme = useTheme();
 
   // Negative Numbers should be zero, percentages over 100 should be capped
+  // A zero denominator should show as 0% fill
   num < 0 ? (num = 0) : { num };
-  const fraction = Math.min((num / den) * 100, 100);
+  const fraction = den <= 0 ? 0 : Math.min((num / den) * 100, 100);
```
hlminh2000, line 15: technically I think it's better to do `den === 0`, because negative denominator can give a very valid fraction :P Wrapping the whole thing in `Math.abs` will account for both ends of the number line :)

but this should work just fine until with we have such usecase :)

anncatton: yeah i decided on this version, i was thinking if there was a negative denominator we might still want to have it displayed as 0. But atm i can't think of a scenario that would create that :P

PR #1399, opened 2020-04-29
Modify Facet Styles and Behaviour

wajiha-oicr:
**Description of changes**

- Renamed and moved IntegerField Component + story into Facet Folder, called it NumberRangeFacet 
- Made multiple requested styling adjustments for the Facet component and the NumberRangeFacet Component
- Adjustment to the facet layout in the file repo page ( allow for both types of facet )
- Added ability for selectAll button to work on queried search params ( for multiple searches )
- Removed regex that was restricting the number field input, deferred to an external validation check that runs when input is out of focus ( `onBlur` ). Decimals now allowed.
- Added memo wrapper on search field regex
- Added a few more mock examples just to get a better picture because why not

**Type of Change**

- [x] Bug
- [x] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


uikit/OptionsList/index.tsx
```diff
@@ -137,30 +155,59 @@ const OptionsList: React.ComponentType<{
...
+          <div
+            css={css`
+              display: flex;
+              justify-content: flex-end;
+              padding: 4px 12px 0px 12px;
+            `}
+          >
+            <Typography variant={'caption'} color={theme.colors.grey}>
+              # files
+            </Typography>
```
hlminh2000, line 167: `files` is only specific to the file repo page, let's extract this out to be an optional prop for the component :) could be something like `countUnit` or something like that and only render the whole block if something is provided :)

uikit/form/Checkbox/index.tsx
```diff
@@ -100,9 +103,12 @@ const Checkbox = ({
   'aria-label': ariaLabel,
   value,
   color,
+  small,
+  ...props
 }: {
   checked: boolean;
   disabled?: boolean;
+  small?: boolean;
```
hlminh2000, line 111: rather than a `small` boolean prop, what do you think about a `size` enum prop that's `"SM" | "MD"` with a `"MD"` default? that way we can support new standard sizes in the future without adding on more props :)

PR #1408, opened 2020-04-30
✨ Add stats bar, icons

anncatton:
**Description of changes**

Adds StatsBar component to File Repository
- displayed with mock data
- new icons: `crosshairs`, `filesize`

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


components/pages/file-repository/StatItem/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { capitalize } from 'global/utils/stringUtils';
+
+type TSTatType = 'file' | 'primary site' | 'donor' | 'program' | 'filesize';
+type StatItemProps = {
+  iconName: UikitIconNames;
+  statType: TSTatType;
+  count: number;
+};
+
+// for mock data display, not sure how this will be calculated esp for file size
```
wajiha-oicr, line 16: I'm not entirely sure either how the file size data is going to look like when it comes in, but when formatting that in the donor table, I based it under the assumption the information would arrive in bytes ( this makes sorting easier :P ). Then I imported and used `fileSize` which automatically handles formatting the number with the correct unit (i.e kb, mb, tb etc) . 


https://www.npmjs.com/package/filesize
https://github.com/icgc-argo/platform-ui/blob/develop/components/pages/file-repository/FileTable/index.tsx#L107

anncatton: Cool! Yeah I wasn't sure either, so i put some oversimplified math

components/pages/file-repository/StatsCard/StatItem.tsx
```diff
@@ -0,0 +1,54 @@
+import filesize from 'filesize';
+import { css } from 'uikit';
+import { UikitIconNames } from 'uikit/Icon/icons';
+import { capitalize } from 'global/utils/stringUtils';
+import { Col } from 'react-grid-system';
+import { useTheme } from 'uikit/ThemeProvider';
+import Icon from 'uikit/Icon';
+import Typography from 'uikit/Typography';
+
+type TSTatType = 'file' | 'primary site' | 'donor' | 'program' | 'filesize';
```
wajiha-oicr, line 10: Just curious about the naming convention here? What does TSTatType mean?

components/pages/file-repository/StatsCard/StatItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type StatItemProps = {
+  iconName: UikitIconNames;
+  statType: TSTatType;
+  count: number;
+};
+
+// for mock data display. this could change when data is integrated
+const getDisplayStat = (type: TSTatType, count: number): string => {
+  return type === 'filesize' ? `${filesize(count)}` : `${count} ${capitalize(type)}s`;
```
wajiha-oicr, line 20: 👍👍👍

Also, adding 's' works but if there's ever a case where there's only 1 of them then the grammar doesn't make sense. I think `pluralize` can take care of the grammar rule in these cases. Not needed for now, but something that may come up later.

components/pages/file-repository/StatsCard/StatItem.tsx
hlminh2000: there is a naming strategy where you prefix the type of the thing with a shorthand for that type, so `T_StatType` to denote a type, `I_StatType` to denote an interface, `C_StatType` to denote a class...
Kinda an old school convention that's more helpful if we were using raw vim in my opinion :P

For us I think it's more noisy than helpful so wouldn't recommend unless there's a very tricky namespace collision going on, even then would say it's a last resort :P

anncatton: Also noticed I had an extra capital `T` in there so will change that too :)

hlminh2000: lol I totally thought you were going for the type prefix :P

anncatton: oh yeah the first `T` was that for sure...but then I capitalized the next one too cuz I can't type :)

components/pages/file-repository/StatsCard/StatItem.tsx
anncatton: Yeah, I don't think the case will come up, but i'll add it in for safety :)

components/pages/file-repository/StatsCard/StatItem.tsx
hlminh2000: lol didn't even notice that :P like Jon says, I can't read...

PR #1420, opened 2020-05-04
Feat/clinical timeline invalid

ciaranschutte:
**Description of changes**

adds invalid tag to timeline tabs
same icon can be used for table when we know the data shape

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/donor/ClinicalTimeline/timeline.tsx
+++ b/components/pages/submission-system/donor/ClinicalTimeline/timeline.tsx
@@ -5,6 +5,8 @@ import { getTimelineStyles } from './util';
 import useTheme from 'uikit/utils/useTheme';
 import VerticalTabs from 'uikit/VerticalTabs';
 import { Entity, EntityType } from './types';
+import Icon from 'uikit/Icon';
+import { InvalidIcon } from './common';
 
 const DayCount = ({ days }: { days: number }) => (
   <div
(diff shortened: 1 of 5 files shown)


components/pages/submission-system/donor/ClinicalTimeline/timeline.tsx
```diff
@@ -27,14 +29,14 @@ const DayCount = ({ days }: { days: number }) => (
 );
 
 type TimeLineItemProps = {
-  item: { id: string; description: string; type: string };
+  item: Pick<Entity, 'id' | 'description' | 'type' | 'invalid'>;
```
hlminh2000, line 32: any harm from just giving it all of `Entity`?

ciaranschutte: @hlminh2000 just not the type. The visual component prop `item` has no need for `data` or `samples` or `interval`

ciaranschutte: no harm, but not "correct" either surely?

hlminh2000: If the purpose is to keep this component unaware of domain data, it probably shouldn't import `Entity` and `EntityType` to begin with. Otherwise, it's already domain aware, the abstraction isn't necessary.

That said, I don't have a strong opinion about this, just thought it could help simplify things :)

PR #1465, opened 2020-05-19
Maintenance page static build

hlminh2000:
- refactors the `pages.tsx` file to contain only internal pages
- remove links to internal pages in the maintenance page
- improve responsiveness for the error pages
- adds some setup to export the maintenance page as a static web bundle.


hlminh2000:
of course I caused some cornflakes...

PR #1472, opened 2020-05-25
✨1440 - init filters context

anncatton:
**Description of changes**

Adds FiltersContext to File Repository
Integrates filters with url state
Uses mock filter button and sqon values for testing, will be connected to facets/search in next PR
Small change to NavBar login redirect to fix bug that prevented login from repository with filters in url

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


global/hooks/useFiltersContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+  setFilters: (filter: FiltersType) => null,
+});
+
+export function FiltersProvider({ children }) {
+  const router = useRouter();
+  const currentQuery = getParams(router);
+
+  const [filters, setCurrentFilters] = React.useState(
+    JSON.parse(currentQuery.filters || null) || defaultFilters,
+  );
```
hlminh2000, line 75: why is this state necessary? `useUrlParamState` already provides a react state binding so we can just operate on that directly can't we?

global/hooks/useFiltersContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+export type FiltersType = {
+  op: CombinationKeys;
+  content: Operator[];
+};
+
+type FiltersContextType = {
+  filters: FiltersType;
+  clearFilters: () => void;
+  setFilters: (FiltersType) => void;
+};
```
hlminh2000, line 59: is this stuff accessible from `sqon-builder`? looks like we're really just interested in `FiltersType` eh

anncatton: Yeah we would just need to complete type...but the types aren't exported atm. Would that be worth a change in `sqon-builder`?

hlminh2000: I see :P yeah it'd be a nice thing, but not necessary right now :) just wanted to check

anncatton:
> I don't think the file `global/hooks/useFiltersContext.tsx` is really meant for application-wide reuse eh, might want to use have it defined somewhere in `components/pages/file-repository` cause it's really only relevant for the file repo

@rosibaj said the filters may be carried through to some other pages in future, like we have on some of the gdc entity pages. So maybe keep in global?

rosibaj:
> > I don't think the file `global/hooks/useFiltersContext.tsx` is really meant for application-wide reuse eh, might want to use have it defined somewhere in `components/pages/file-repository` cause it's really only relevant for the file repo
> 
> @rosibaj said the filters may be carried through to some other pages in future, like we have on some of the gdc entity pages. So maybe keep in global?

Happy to have call to discuss future business cases if needed!

global/hooks/useFiltersContext.tsx
anncatton: yes i misunderstood all the things the param state was providing! reinvented the wheel a bit :)

PR #1478, opened 2020-05-26
Feat/#1426 program services modal

ciaranschutte:
**Description of changes**

Add program services modal

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


global/config.ts
```diff
@@ -17,8 +17,8 @@ export const getConfig = () => {
       publicConfig.EGO_API_ROOT || '',
       `/api/oauth/login/google?client_id=${publicConfig.EGO_CLIENT_ID || ''}`,
     ),
-    DOCS_URL_ROOT: publicConfig.DOCS_URL_ROOT || 'https://docs.icgc-argo.org/docs/',
     DACO_URL: publicConfig.DACO_URL || 'https://icgc.org/daco',
+    DOCS_URL_ROOT: publicConfig.DOCS_URL_ROOT || 'https://docs.icgc-argo.org',
```
hlminh2000, line 21: this looks ok to change for me, but wanna confirm... is anything existing affected by this?

components/pages/Homepage/index.tsx
```diff
@@ -19,6 +19,8 @@ import {
   ResponsiveGridLayout,
 } from './common';
 import { FILE_REPOSITORY_PATH } from 'global/constants/pages';
+import { ModalPortal } from 'components/ApplicationRoot';
+import ProgramServicesModal from './ProgramServicesModal';
```
hlminh2000, line 23: these aren't used here eh?

components/NavBar.tsx
```diff
@@ -154,11 +163,12 @@ export default function Navbar({ hideLink = false, disableLogoLink = false }) {
...
-                    <Typography variant={'default'}>Submission</Typography>
-                  </MenuItem>
+                  {' '}
+                  {programServices}
                 </a>
               </Link>
+            ) : (
+              <div onClick={() => setModalVisibility(!isModalVisible)}> {programServices}</div>
             )}
             {!userModel && (
```
anncatton, line 173: I think the login menu item needs some extra margin now that program services is showing for logged out state

global/config.ts
ciaranschutte: I don't think so. 
Docs pages are linked here: https://github.com/icgc-argo/platform-ui/blob/develop/global/constants/docSitePaths.tsx with the `/docs/` prefix in their strings

The root for the text documentation, not dictionary, does not have `/docs/` in it's path @hlminh2000

components/NavBar.tsx
ciaranschutte: @anncatton nice catch - thank you! Added horizontal padding to login button menu item

PR #1524, opened 2020-06-02
Hide privacy page, switch to external link

wajiha-oicr:
**Description of changes**

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


.env.schema
```diff
@@ -3,6 +3,7 @@ NODE_ENV=development | production
 AUTH_DISABLED=false
 FEATURE_REPOSITORY_ENABLED=false
 FEATURE_LANDING_PAGE_STATS_ENABLED=false
+FEATURE_PRIVACY_PAGE_ENABLED=false
```
hlminh2000, line 6: we're not going to maintain this page moving forward so the whole thing can be removed ;)

wajiha-oicr: wow... rest in peace, privacy page

global/constants/icgcPaths.tsx
```diff
@@ -0,0 +1,5 @@
+export const ICGC_ROOT = 'https://www.icgc-argo.org';
+
+export const ICGC_PRIVACY_PATH = `${ICGC_ROOT}/page/2/privacy`;
+export const ICGC_TERMS_PATH = `${ICGC_ROOT}/page/1/terms-and-conditions`;
+export const ICGC_PUBLICATION_PATH = `${ICGC_ROOT}/page/77/e3-publication-policy`;
```
hlminh2000, line 5: not a big deal but I'd suggest naming these `ARGO_PRIVACY_PAGE` or something like that, cause `path` technically means just the `/...` portion of the url. Also `ARGO` to differentiate from original icgc

hlminh2000: maybe the file name can be updated too

PR #1571, opened 2020-06-11
stats bar data hookup

hlminh2000:
**Description of changes**

- primary sites is left out as not currently available

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/file-repository/StatsCard/STATS_BAR.gql
+++ b/components/pages/file-repository/StatsCard/STATS_BAR.gql
@@ -3,7 +3,7 @@ query STATS($filters: JSON) {
     hits(filters: $filters) {
       total
     }
-    aggregations(filters: $filters, include_missing: true) {
+    aggregations(filters: $filters, include_missing: true, aggregations_filter_themselves: true) {
       donors__donor_id {
         buckets {
           key
--- a/components/pages/file-repository/StatsCard/STATS_BAR.gql
+++ b/components/pages/file-repository/StatsCard/STATS_BAR.gql
@@ -1,0 +1,24 @@
+query STATS($filters: JSON) {
+  file {
+    hits(filters: $filters) {
+      total
+    }
+    aggregations(filters: $filters, include_missing: true) {
+      donors__donor_id {
+        buckets {
+          key
+        }
+      }
+      study_id {
+        buckets {
+          key
+        }
+      }
+      file__size {
+        stats {
+          sum
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


components/pages/file-repository/StatsCard/index.tsx
```diff
@@ -21,17 +21,54 @@ import { css } from 'uikit';
...
+  const { filters } = useFiltersContext();
+
+  const { data, loading } = useQuery<StatsBarQueryResponseData, StatsBarQueryInput>(STATS_BAR, {
+    variables: {
+      filters,
+    },
+  });
+
+  const filesCount = data ? data.file.hits.total : 0;
+  const donorsCount = data ? data.file.aggregations.donors__donor_id.buckets.length : 0;
```
anncatton, line 67: Do we want any of these agg fields to come from enums?

components/pages/file-repository/StatsCard/StatItem.tsx
```diff
@@ -27,24 +27,22 @@ import { useTheme } from 'uikit/ThemeProvider';
 import Icon from 'uikit/Icon';
 import Typography from 'uikit/Typography';
 
-type StatType = 'file' | 'primary site' | 'donor' | 'program' | 'filesize';
+type StatType = 'file' | 'primary site' | 'donor' | 'program' | 'fileSize';
```
anncatton, line 30: i guess no reason to modify `primary site` since i think we'll need to remove it?

components/pages/file-repository/StatsCard/STATS_BAR.gql
```diff
@@ -0,0 +1,24 @@
+query STATS($filters: JSON) {
+  file {
+    hits(filters: $filters) {
+      total
+    }
+    aggregations(filters: $filters, include_missing: true) {
```
anncatton, line 6: do we need the `aggregations_filter_themselves` here?

hlminh2000: yes we sure do! thanks for the catch :)

components/pages/file-repository/StatsCard/StatItem.tsx
hlminh2000: I'll talk to Rosi to figure out what we want to do about this one. Will probably address update in a later PR :)

components/pages/file-repository/StatsCard/index.tsx
hlminh2000: hmmm I don't think it's necessary for the fields in the graphql schema. We have a CI test that validates the gql queries against the api in dev, so invalid queries will fail CI. The TS types are already there to prevent mistakes in this particular component.

On a separate note though, and more for the gateway, I should try to find a way to perform the same query validation test from the gateway's CI pipeline to prevent breaking changes from there

PR #1615, opened 2020-06-17
filerepo facets smoother transitions

jaserud:
**Description of changes**

- keep facet aggregations in a state to render between filter change loads
- fade facet container and disable during loading 
    - DNALoader didn't disable and cover entire facet container, can still see container when you scroll down on it

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/file-repository/FacetPanel/index.tsx
+++ b/components/pages/file-repository/FacetPanel/index.tsx
@@ -235,7 +235,7 @@ export default () => {
 
   const [aggregations, setAggregations] = React.useState({});
 
-  const { data, loading = true } = useFileFacetQuery(filters, {
+  const { data, loading } = useFileFacetQuery(filters, {
     onCompleted: (data) => {
       setAggregations(data ? data.file.aggregations : {});
     },
--- a/components/pages/file-repository/FacetPanel/index.tsx
+++ b/components/pages/file-repository/FacetPanel/index.tsx
@@ -233,21 +233,13 @@ export default () => {
   const theme = useTheme();
   const { filters, setFilterFromFieldAndValue, replaceAllFilters } = useFiltersContext();
 
-  const { data, loading } = useFileFacetQuery(filters);
+  const [aggregations, setAggregations] = React.useState({});
 
-  const [aggregations, setAggregations] = React.useState(data ? data.file.aggregations : {});
-
-  const [isFacetLoading, setIsFacetLoading] = React.useState(true);
-
-  React.useEffect(() => {
-    if (loading) {
-      setIsFacetLoading(true);
-      return;
-    }
-
-    setAggregations(data ? data.file.aggregations : aggregations);
-    setIsFacetLoading(false);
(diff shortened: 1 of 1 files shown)


components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -233,8 +233,13 @@ export default () => {
...
 
-  const { data, loading } = useFileFacetQuery(filters);
-  const aggregations = data ? data.file.aggregations : {};
+  const [aggregations, setAggregations] = React.useState({});
+
+  const { data, loading = true } = useFileFacetQuery(filters, {
+    onCompleted: (data) => {
+      setAggregations(data ? data.file.aggregations : {});
+    },
+  });
```
hlminh2000, line 242: is this actually necessary? `loading` being `true` doesn't change `data` does it?

jaserud: No dont think it does. I just wanted the FacetContainer to start faded and then become enabled. Some times on my local this variable starts with false... not a big thing I can change it it if we prefer

hlminh2000: yeah maybe try testing it without this and just hard refresh, it'd be cleaner if we didn't have to handle another state

jaserud: Works without it so removed :+1:

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -233,8 +233,13 @@ export default () => {
   const theme = useTheme();
   const { filters, setFilterFromFieldAndValue, replaceAllFilters } = useFiltersContext();
 
-  const { data, loading } = useFileFacetQuery(filters);
-  const aggregations = data ? data.file.aggregations : {};
+  const [aggregations, setAggregations] = React.useState({});
```
hlminh2000, line 236: why do we still need this?

jaserud: I'm using the state to remember the aggreagations set for the facets...From what I can see the FacetContainer blips from empty to updated because it has no data while the query is loading. So I figured we could just hold on to it in a state and update once the query is done

hlminh2000: hmmm it's surprising that that's happening, we never had to do this. If this was normal apollo behaviour we'd have code like this all over, but we don't... so I feel like something else is going on here...

anncatton: is the way i've done the data hookup maybe not working with what hte containers/loader state expects?

hlminh2000: no I think what we have there is fine @anncatton, I'm taking a look at this right now and I'm tracing things back to the `useUrlParamState` thing, something might be double firing there...

PR #1622, opened 2020-06-18
Add error icons and notification banner to donor summary section

wajiha-oicr:
**Description of changes**
- adds type for the `validWithCurrentDictionary` that comes with each donor summary record,
uses it to display an error icon in the table
- also uses it to manually count the total invalid donors, but expecting that to come as a statistic later ( like the other statistics )

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
+++ b/components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
@@ -74,7 +74,7 @@ export default ({
     fullyReleasedDonorsCount: 0,
     partiallyReleasedDonorsCount: 0,
     noReleaseDonorsCount: 0,
-    invalidDonorsCount: 0,
+    donorsInvalidWithCurrentDictionaryCount: 0,
   };
 
   const containerRef = createRef<HTMLDivElement>();
@@ -385,12 +385,7 @@ export default ({
             css={css`
               opacity: ${isTableLoading ? 0.5 : 1};
             `}
-            programDonorSummaryStats={{
-              ...programDonorSummaryStats,
-              invalidDonorsCount: programDonorSummaryEntries.filter(
-                (donor) => !donor.validWithCurrentDictionary,
-              ).length,
-            }}
+            programDonorSummaryStats={programDonorSummaryStats}
           />
         }
         noMargin={true}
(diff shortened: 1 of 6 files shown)


components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
```diff
@@ -382,7 +385,12 @@ export default ({
             css={css`
               opacity: ${isTableLoading ? 0.5 : 1};
             `}
-            programDonorSummaryStats={programDonorSummaryStats}
+            programDonorSummaryStats={{
+              ...programDonorSummaryStats,
+              invalidDonorsCount: programDonorSummaryEntries.filter(
+                (donor) => !donor.validWithCurrentDictionary,
```
jaserud, line 391: are we not using `invalidDonorsCount` from the `programDonorSummaryStats` because it returns zero?? `programDonorSummaryEntries` is only the entries in the current table page so you will not get all `invalidDonorsCount`. Or is there more to this that I don't know about

components/pages/submission-system/program-dashboard/DonorDataSummary/index.tsx
```diff
@@ -94,6 +95,12 @@ export default () => {
   const isDonorSummaryEntriesEmpty =
     !programDonorSummaryStats || programDonorSummaryStats.registeredDonorsCount === 0;
 
+  const numInvalidDonors = programDonorSummaryEntries.filter(
+    (donor) => !donor.validWithCurrentDictionary,
+  ).length;
+
+  // todo: instead programDonorSummaryStats.invalidDonorsCount instead, once ready
```
jaserud, line 102: Same question as before. I guess I'm trying to understand why not just use it and when its implemented in gateway it'll just work instead of having to come back to fix this again :thinking: ??

hlminh2000:
@jaserud @wajiha-oicr I've added this aggregation on the gateway, PR pending here: https://github.com/icgc-argo/platform-api/pull/274. 

Once this is merged, please update to use this aggregation as the client-side aggregation will not be complete :)

components/pages/submission-system/program-dashboard/DonorDataSummary/index.tsx
jaserud: nvm Minh's gateway PR has answered my question :+1:

components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
jaserud: nvm, I see that this value wasn't available and Minh's gateway PR is adding that

PR #1690, opened 2020-07-06
💄 1651 - Width styling for mobile modals

anncatton:
**Description of changes**

Fixes modal view on mobile.

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


components/ApplicationRoot.tsx
```diff
@@ -233,6 +231,7 @@ export default function ApplicationRoot({
                     left: 0px;
                     top: 0px;
                     z-index: 9999;
+                    width: -webkit-fill-available;
```
hlminh2000, line 234: this will leave out non-webkit based browsers. I think we can make a couple of sharable css blocks in uikit that includes a bunch of vendor specific versions of the same thing bundled together and reuse around so it's consistently browser-agnostic.

A little verbose but helps maintain browser compatibility so worth the cost I think :)

uikit/Modal/index.tsx
```diff
@@ -66,11 +68,14 @@ const ModalOverlay = styled('div')`
...
-      .hsl()
-      .string()};
+  background: ${({ theme }) => Color(theme.colors.primary_dark).alpha(0.8).hsl().string()};
+  height: -webkit-fill-available;
+  width: -webkit-fill-available;
+  min-height: -webkit-fill-available;
+  @media (min-width: 768px) {
+    width: 100vw;
+    height: 100vh;
+  }
```
hlminh2000, line 78: just curious what problem is this addressing?

components/ApplicationRoot.tsx
hlminh2000: something like this:
```
const fillWidth = css`
  width: -webkit-fill-available;
  width: -moz-available;
`
```

uikit/Modal/index.tsx
anncatton: Apparently viewport calculations don't work properly on mobile - which was causing the modal to position on the side - so I set this calculation to work only on larger screens.

PR #1703, opened 2020-07-08
✨ 1666 - Add SystemAlert component

anncatton:
**Description of changes**

Creates SystemAlert component (no integration to platform)
Assumes an api response for notifications: `[{ level: AlertLevel; title: string; message?: string; dismissable: boolean; }, ...]`

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/SystemAlert/index.tsx
+++ b/uikit/SystemAlert/index.tsx
@@ -1,0 +1,126 @@
+import React from 'react';
+import { ThemeColorNames } from '../theme/types';
+import theme from 'uikit/theme/defaultTheme';
+import Typography from 'uikit/Typography';
+import Icon from 'uikit/Icon';
+import { UikitIconNames } from 'uikit/Icon/icons';
+import { css } from 'uikit';
+import { Col, Row } from 'react-grid-system';
+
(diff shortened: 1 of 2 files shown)


uikit/SystemAlert/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+type AlertVariant = {
+  [key in AlertLevel]: {
+    color: keyof ThemeColorNames;
+    icon: UikitIconNames;
+    fill: keyof ThemeColorNames;
+  };
+};
+const ALERT_VARIANTS: AlertVariant = {
+  error: {
+    color: theme.colors.error as keyof ThemeColorNames,
```
hlminh2000, line 21: let's `defaultTheme` is good for just grabbing the types but let's not use its values directly. Prefer accessing theme through the `useTheme` hook so it can be swap out in run time if that becomes a need. Could turn this into a function that accepts `theme: typeof defaultTheme` if you want to keep this outside of the component :)

PR #1717, opened 2020-07-21
build responsive logic for navbar

wajiha-oicr:
**Description of changes**

- Made a NavElement type and component so that navigation links ( or if it isn't a link but has click behaviour) could be reused between different layouts
- Rewrote the existing links ( file repository, program services, profile, logout) as the new NavElement object types, kept in arrays
- Use state logic for the mobile dropdown visibility, and general mobile layout toggle
- Used "medium" screen as breakpoint for mobile layout
- Styling adjustments to support responsiveness

**Type of Change**

- [ ] Bug
- [x] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour


ciaranschutte:
hover
- green hover colour is very hard to make out because the dropdown is using a lighter background than top header bar (i've mentioned Kim in a comment on the issue)

opendouble
- open full menu, resize, open mobile menu, resize to full. there will be a double menu

ciaranschutte:
@wajiha-oicr @hlminh2000 If we're starting to work on responsiveness should we use Emotion media queries to reduce state in components? https://emotion.sh/docs/media-queries

wajiha-oicr:
> https://emotion.sh/docs/media-queries



> @wajiha-oicr @hlminh2000 If we're starting to work on responsiveness should we use Emotion media queries to reduce state in components? https://emotion.sh/docs/media-queries

being able to use media queries in emotion's css is pretty awesome 😎. Just wondering what component to apply it to here and with what kind of rule? I feel like because the mobile version ends up changing a few significant things about the layout, like the order of the buttons and the components that are used, I wouldn't know how to do those with purely css rules?

hlminh2000:
Media queries are great for addressing very specific problems that cannot be (or is clunky) to do otherwise. The problem is it tends to add a lot of "magic" values in the code that are hard to see the relationship between (random screen sizes), so I'd still like to work with named screen size classes (xs, sm, md, ...) generally and use media queries for edge case problems that's not captured by screen classes. 

`react-grid-system` has a `useScreenClass` hook so I'd start with using that first and address very specific problems that doesn't fit into that model with media queries.

components/NavBar.tsx
```diff
@@ -126,12 +129,97 @@ export default function Navbar({ hideLinks = false, disableLogoLink = false }) {
     }
   }, [path, query]);
 
-  const programServices = (
-    <MenuItem ref={React.createRef()} active={path.search(SUBMISSION_PATH) === 0 || isModalVisible}>
-      <Typography variant={'default'}>Program Services</Typography>
-    </MenuItem>
+  React.useEffect(() => {
+    // required for inital page load, window object accessible in useffect only
+    setMobileLayout(window.innerWidth <= reactGridBreakpoints.md);
```
wajiha-oicr, line 134: @hlminh2000 would you suggest using the screenclass hook here to control the state of the isMobileLayout boolean ?

hlminh2000: Yes :) I was just typing that up lol

wajiha-oicr: okay I'd love to see what you have as a suggestion! thanks

hlminh2000: oh sorry I meant exactly what you're thinking (I think you already have the idea)
something like this?
```
const screenClass = useScreenClass()
const isMobileLayout = ["xs", "sm", "md"].includes(screenClass)
```

wajiha-oicr:
@ciaranschutte @hlminh2000 changes made! please take a look when you have a chance:) :)

package.json
```diff
@@ -89,7 +89,7 @@
     "react-dom": "^16.13.1",
     "react-ga": "^2.6.0",
     "react-google-recaptcha": "^2.0.1",
-    "react-grid-system": "^4.4.6",
```
hlminh2000, line 92: this is a major version update, which is theoretically a breaking change... we should do a round of inspection around every page to make sure nothing is broken

wajiha-oicr: Actually, I just noticed there's a breaking change within this very file itself! Relying on the useScreenClass Hook breaks the initial render of the user profile component in mobile view? Super strange I'm scratching my head as to why. 


image


It fixes itself upon re-size but this wasn't happening with the previous resize detection method.

wajiha-oicr: perhaps it isn't playing well with server side rendering because under the hood it's relying on the (unavailable?) window object? any ideas @hlminh2000?

hlminh2000: @wajiha-oicr this seems to have fixed it for me. https://github.com/icgc-argo/platform-ui/pull/1722
Still not sure what the problem is but it seems to fix it :/

wajiha-oicr: just tested it out and it works! thanks Minh, merged!

PR #1720, opened 2020-07-27
redo systemAlert with flexbox, add new knobs

wajiha-oicr:
**Description of changes**
- added some new knobs to the system Alert story
- the icon needs to stay affixed to the left of the message at a constant distance, changed to a flexbox layout to achieve this

**Type of Change**

- [ ] Bug
- [x] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/uikit/SystemAlert/index.tsx
+++ b/uikit/SystemAlert/index.tsx
@@ -51,60 +51,56 @@ const SystemAlert: React.ComponentType<AlertProps> = ({ alert, onClose }) => {
   const theme = useTheme();
   const icon = ALERT_VARIANTS[alert.level].icon;
   return (
-    <Row
+    <div
       css={css`
         background-color: ${theme.colors[ALERT_VARIANTS[alert.level].color]};
-        padding: 10px 0;
+        padding: 12px;
+        display: flex;
+        justify-content: space-between;
+        align-items: flex-start;
       `}
     >
-      <Col
-        sm={1}
+      <div
         css={css`
           display: flex;
-          justify-content: center;
-          align-items: center;
         `}
       >
-        <Icon name={icon} width="30px" height="30px" />
-      </Col>
-      <Col
-        sm={10}
-        css={css`
-          display: flex;
-          flex-direction: column;
-          justify-content: center;
-        `}
-      >
-        <Typography
-          css={css`
-            font-size: 16px;
-            font-weight: 500;
-            color: ${theme.colors[ALERT_VARIANTS[alert.level].fill]};
-            margin: 0;
-          `}
-        >
-          {alert.title}
-        </Typography>
-        {alert.message && (
+        <div>
+          <Icon
+            name={icon}
+            width="30px"
+            height="30px"
+            css={css`
+              margin-right: 15px;
+            `}
+          />
+        </div>
+        <div>
           <Typography
             css={css`
-              margin: 0;
+              font-size: 16px;
+              font-weight: 500;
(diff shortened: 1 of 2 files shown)


uikit/SystemAlert/stories.tsx
```diff
@@ -1,35 +1,56 @@
...
 
+const getNewMessage = {
+  original: (input: string) => input,
+  short: (input: string) => input?.substring(0, 10),
+  long: (
+    input: string,
+  ) => `${input}. Here is an example of a large message body. Bitters vinyl VHS blog dreamcatcher ramps.
+  Readymade squid air plant, chicharrones jean shorts kitsch meggings occupy next level fixie 
+  cold-pressed vegan fam intelligentsia gluten-free. Flannel next level schlitz try-hard. Bespoke 
+  kickstarter edison bulb, street art taxidermy iPhone pitchfork woke fashion axe leggings af sartorial fanny pack.`,
```
hlminh2000, line 34: hipster ipsum never gets old lol

uikit/SystemAlert/stories.tsx
```diff
@@ -1,35 +1,56 @@
...
   const [currentAlert, setCurrentAlert] = React.useState(defaultAlert);
   return currentAlert ? (
-    <SystemAlert alert={currentAlert} onClose={() => setCurrentAlert(null)} />
+    <SystemAlert
+      alert={{
+        ...currentAlert,
+        dismissable,
+        message: getNewMessage[messageSize](currentAlert.message),
+      }}
+      onClose={() => setCurrentAlert(null)}
```
hlminh2000, line 52: does `currentAlert` actually ever get set?

uikit/SystemAlert/index.tsx
```diff
@@ -51,60 +51,56 @@ const SystemAlert: React.ComponentType<AlertProps> = ({ alert, onClose }) => {
...
-        {alert.message && (
+        <div>
+          <Icon
+            name={icon}
+            width="30px"
+            height="30px"
+            css={css`
+              margin-right: 15px;
+            `}
+          />
```
hlminh2000, line 76: icons have a `fill` prop for setting the color eh https://1710-bannerupdates--argo-ui-storybook.netlify.app/?path=/story/uikit-icon--basic

uikit/SystemAlert/stories.tsx
wajiha-oicr: I believe it's being set in its declaration by the defaultAlert value, which comes from the knobs (line 30) https://github.com/icgc-argo/platform-ui/pull/1720/files#diff-83eefd21997928f1e5da1183bfb7d8d0L30

uikit/SystemAlert/stories.tsx
wajiha-oicr: LOL they even have an API you can call hahah!

uikit/SystemAlert/index.tsx
wajiha-oicr: that's correct, but the icons are all defaulting as black fill which is the color for now.

Actually, these particular icons are kind of tricky because there are two path definitions, but it's the internal path that needs to have a transparent fill and the outer one needs to have the fill applied. Though I think the current way the mask is being applied, when the inner layer is transparent it still shows the outer layer rather than a cut-out.

hlminh2000: I see, yeah we probably need some new icons from Kim and just add to the collection eh :)

PR #1738, opened 2020-08-26
Add Page Titles Throughout site

wajiha-oicr:
**Description of changes**

- Subtitles added to ICGC ARGO for various site pages

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/head.tsx
+++ b/components/pages/head.tsx
@@ -25,10 +25,12 @@ const defaultDescription = '';
 const defaultOGURL = '';
 const defaultOGImage = '';
 
-const Head = props => (
+const Head = (props) => (
   <NextHead>
     <meta charSet="UTF-8" />
-    <title>{props.title || 'ICGC ARGO'}</title>
+    <title>
+      {(props.subtitle && `ICGC ARGO | ${props.subtitle}`) || props.title || 'ICGC ARGO'}
+    </title>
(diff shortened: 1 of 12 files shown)


components/pages/head.tsx
```diff
@@ -45,6 +47,7 @@ const Head = props => (
 
 Head.propTypes = {
   title: string,
+  subtitle: string,
   description: string,
   url: string,
   ogImage: string,
```
hlminh2000, line 53: A small favour to ask :) let's just move this to typescript :P

pages/repository/index.tsx
```diff
@@ -33,5 +33,5 @@ export default createPage<{ egoJwt: string }>({
     }
   },
 })(() => {
-  return <FileRepositoryPage />;
+  return <FileRepositoryPage subtitle="File Repository" />;
```
hlminh2000, line 36: why don't we just have this in the `components/pages/file-repository/index.tsx` like the other pages?

wajiha-oicr: no good reason, was trying to match the original convention for this one in case there was some reasoning for subtitle to be an exposed prop. though I can't think of any cases why, so I agree it's best to match the convention of the others 👍🏽

PR #1752, opened 2020-09-25
add international to the country code validation

wajiha-oicr:
**Description of changes**

- let "INTL" be a valid country code when making a shortname

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/program-form/validations.tsx
+++ b/components/pages/submission-system/program-form/validations.tsx
@@ -30,7 +30,10 @@ const baseValidations = yup.object().shape({
       const providedCode = value.slice ? value.slice(-2) : null;
       return !!COUNTRIES.find((country) => country.code === providedCode);
     })
-    .matches(/-([A-Z][A-Z])$/, '${label} must end with a 2 character country code: "-XX"')
+    .matches(
+      /-([A-Z][A-Z]|INTL)$/,
+      '${label} must end with a 2 character country code: "-XX", or end with "-INTL" for international programs.',
+    )
     .matches(/^[A-Z0-9]/, '${label} must begin with an uppercase letter or a number')
     .matches(
       /^[A-Z0-9_-]+$/,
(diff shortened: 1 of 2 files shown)


components/pages/submission-system/program-form/validations.tsx
```diff
@@ -30,7 +30,10 @@ const baseValidations = yup.object().shape({
       const providedCode = value.slice ? value.slice(-2) : null;
       return !!COUNTRIES.find((country) => country.code === providedCode);
     })
-    .matches(/-([A-Z][A-Z])$/, '${label} must end with a 2 character country code: "-XX"')
+    .matches(
+      /-([A-Z][A-Z]|INTL)$/,
+      '${label} must end with a 2 character country code: "-XX", or end with "-INTL" for international programs.',
+    )
```
hlminh2000, line 36: :/ if we are already checking against a hard-coded list, why is this regex check still necessary? :/

hlminh2000 (commented):
I know this exists from before, but just curious... if we find out why, maybe a comment would help later :)

components/pages/submission-system/program-form/validations.tsx
wajiha-oicr: I think it helps with checking for the inclusion of a dash ( - ) . I.e. without it, it would be okay with TESTCA instead of TEST-CA

wajiha-oicr: would also prevent someone from reusing the same country code twice. I.e TEST-CACA

hlminh2000: oh I see that makes sense! lol I guess I need coffee...

hlminh2000 (approved):

components/pages/submission-system/program-form/validations.tsx
wajiha-oicr: you're right though, probably a good place to add a comment! I'll add one in to clarify :)

Issue #1737, opened 2020-08-25
🐛 Clicking on Manage Programs a second time breaks display


ciaranschutte:
## Describe the bug
 Clicking on Manage Programs a second time breaks display

## Steps To Reproduce
https://drive.google.com/file/d/1CG9U4sOJbxIT9sNCPjecisTQe2-aVsWM/view

![image](https://user-images.githubusercontent.com/15824252/91312968-d4ca1680-e782-11ea-94a7-7c4d7a6d3fe8.png)

URL parameters are not being passed on the second click

## Expected behaviour
- Manage program data to stay on screen
- Should also work for all clinical submission tabs 
- File repository should also be checked after fix as it has url parameters

ciaranschutte:
@rosibaj required to save state of each page too?
 eg. if on `Manage Programs` => `Profile`, 
clicks on `Manage Program`, 
stays on `Profile` 

 or is normal page reload(state is lost) correct (like https://dcc.icgc.org/search)?

rosibaj:
@ciaranschutte its better to save state.

rosibaj:
@ciaranschutte looks great! i couldnt break it in testing! moving forward.

rosibaj:
@ciaranschutte can i please get a version number added so that we can move this forward?

rosibaj:
confirmed on prod

PR #1754, opened 2020-09-29
URL param hook breaking navigation

ciaranschutte:
**Description of changes**

Add `router.asPath` as dependency for updating tab useEffect
Consistent website behaviour, doesn't save state on reloading the page.

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/global/hooks/useUrlParamState.ts
+++ b/global/hooks/useUrlParamState.ts
@@ -59,13 +59,37 @@ export const useHook = <T>(
   const router = useRouter();
   const currentQuery = getParams(router);
   const [firstRender, setFirstRender] = React.useState(true);
+  const [previousQuery, setPreviousQuery] = React.useState(null);
+  console.log('router', router.asPath);
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


hlminh2000 (approved):

ciaranschutte:
@hlminh2000 checked functionality with Rosi #1737 and we need to preferably retain state.
The entire section still reloads (so we preserve any page level data fetching, hooks etc) but it will go to the previous query value if it has one.

eg. if the `Profile` was the active tab on `Manage Programs` then user reclicks `Manage Program` in sidemenu, it will go to `Profile` tab

global/hooks/useUrlParamState.ts
```diff
@@ -59,14 +46,28 @@ export const useHook = <T>(
   const router = useRouter();
   const currentQuery = getParams(router);
   const [firstRender, setFirstRender] = React.useState(true);
+  const [previousQuery, setPreviousQuery] = React.useState(null);
```
hlminh2000, line 49: can we get some typing for this guy please? :)

PR #1791, opened 2021-01-13
🐛 fonts not showing up in Firefox

samrichca:
Add typography theme info to submenus & buttons so the fonts will render in Firefox.

### DIFF
--- a/components/pages/file-repository/QueryBar/index.tsx
+++ b/components/pages/file-repository/QueryBar/index.tsx
@@ -105,7 +105,7 @@ const content = css`
       cursor: pointer;
       padding: 0 7px;
       margin-right: 6px;
-      font-weight: bold;
+      font-weight: 600;
       cursor: pointer;
     }
     & .sqon-less,
--- a/components/pages/file-repository/QueryBar/index.tsx
+++ b/components/pages/file-repository/QueryBar/index.tsx
@@ -72,6 +72,7 @@ const content = css`
       align-items: center;
       height: 22px;
       border-radius: 8px;
+      font-family: Work Sans, sans-serif;
       font-size: 11px;
       font-weight: 300;
       letter-spacing: 0.2px;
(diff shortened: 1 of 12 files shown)


ciaranschutte (changes requested):
why did package-lock.json update?
If you install with `npm ci` rather than `npm i` it should not touch package-lock

everything else looks great

joneubank:
I don't think package.lock updating is a problem. If our dependency version codes are correctly set, we should be accepting updated versions of packages to get all the latest security fixes and any other improvements without bringing in any breaking changes. In the long run this will help with dependency version management as we'll get incremental changes and not have to deal with massive version changes all at once.

 If we need a specific version of a dependency then it should be specified in the package.json not controlled via `npm ci`. If we need to specify a specific minor version there are version codes for this as well.

An exception to this is any work done as hotfix to master where we won't have the dev/qa testing environments to go through first.

ciaranschutte:
@joneubank Mostly just following on from how Minh and I worked in the sense that we'd collide on package-lock updates a lot which was a pain. 

We also did have a case of a breaking change in a minor patch at one point - even though this "shoudn't" happen

Also means we have to be diligent "If our dependency version codes are correctly set"

joneubank:
Yes, let's be diligent 😉 

I understand these concerns about conflicts. One solution is merging in each other's changes before committing your package.lock changes. Requires a bit more coordination, but with the size of our team it I _think_ it should be ok, but if its causing issues then we can make no package.lock updates without intention our team policy.

Also, I'm aware that NPM has bad version control and that sometimes minor and patch versions of things cause breaking changes. I alluded to this in my comment, its better to find these out incrementally rather than having to organize a "dependencies update" task and find all the breaking things at once. When this happens, we can lock down our version or make fixes, case-by-case.

ciaranschutte:
@joneubank great sounds good moving forward! so for this ticket let's just wait till build passes and approve and get it out the door @samrichca

ciaranschutte (approved):

PR #1809, opened 2021-02-02
✨ add API-driven system alerts

samrichca:
- make a system alerts component
- remove maintenance banner because it's redundant now - it can be added in gateway env variables if needed


joneubank (approved):
Nothing to add. This is very clean and readable, thanks Sam.

PR #1810, opened 2021-02-02
Feat/#1801 access facet feedback

ciaranschutte:
**Description of changes**

- Access facet should be very first facet, pulled facet out of out loop to show before search bar
- Names in querybar should be readable, added more to `displayValue` function
- order of options now fixed

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/file-repository/FacetPanel/index.tsx
+++ b/components/pages/file-repository/FacetPanel/index.tsx
@@ -286,12 +286,6 @@ export default () => {
     return order.map((order) => options.find((option) => option.key === order)).filter(Boolean);
   };
 
-  // function to manipulate options values eg. to preserve order
-  const prepareOptions = (
-    options: FilterOption[],
-    prepare: (options: FilterOption[]) => FilterOption[],
-  ) => prepare(options);
-
   const getOptions: GetAggregationResult = (facet) => {
     return (aggregations[facet.facetPath] || { buckets: [] }).buckets.map((bucket) => ({
       ...bucket,
@@ -465,7 +459,7 @@ export default () => {
           <TooltipFacet
             {...releaseStageCommonProps}
             key={releaseStageDetails.name}
-            options={prepareOptions(getOptions(releaseStageDetails), orderReleaseStage)}
+            options={orderReleaseStage(getOptions(releaseStageDetails))}
             countUnit={'files'}
             onOptionToggle={onFacetOptionToggle(releaseStageDetails)}
             onSelectAllOptions={onFacetSelectAllOptionsToggle(releaseStageDetails)}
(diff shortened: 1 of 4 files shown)


components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -275,6 +276,22 @@ export default () => {
...
+  const orderReleaseStage = (options: FilterOption[]) => {
+    const order = ['OWN_PROGRAM', 'FULL_PROGRAMS', 'ASSOCIATE_PROGRAMS', 'PUBLIC_QUEUE', 'PUBLIC'];
+    return order.map((order) => options.find((option) => option.key === order)).filter(Boolean);
+  };
+
+  // function to manipulate options values eg. to preserve order
+  const prepareOptions = (
+    options: FilterOption[],
+    prepare: (options: FilterOption[]) => FilterOption[],
+  ) => prepare(options);
```
joneubank, line 293: Why make a function that includes the function you want to use as an argument and does nothing else? Especially if the default is a "do nothing" function? Function.

I don't understand the value here, it just seems like an extra layer of complication.

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -416,6 +434,12 @@ export default () => {
     },
   });
 
+  // pull out release stage facet to show before search bar
+  const releaseStageDetails = presetFacets.filter(
+    (f) => f.facetPath === FileFacetPath.release_stage,
+  )[0];
```
joneubank, line 440: This looks like what you want is the `.find` method.
```
const releaseStageDetails = presetFacets.find( f => f.facetPath === FileFacetPath.release_stage);
```

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -534,45 +574,35 @@ export default () => {
           </MenuItem>
         </FacetRow>
         {!loadingFieldDisplayNames &&
-          presetFacets.map((facetDetails) => {
-            const facetProps = commonFacetProps(facetDetails);
+          presetFacets
+            .filter((f) => f.facetPath !== FileFacetPath.release_stage)
```
joneubank, line 578: Maybe add a comment here stating why we are filtering out this particular facet (its obvious if you read the above section, but if you just look here then this code seems odd).

joneubank (changes requested):

components/pages/file-repository/utils/index.ts
```diff
@@ -226,10 +226,20 @@ export const removeFilter: TRemoveFilter = (field, filters) => {
...
+export const toDisplayValue: (value: string, field?: string) => string = (value, field) => {
+  if (value === IS_MISSING) {
+    return 'No Data';
+  }
+
+  if (!!field && !!value) {
+    const displayName = getDisplayName(field, value);
+    return displayName ? displayName : value;
+  } else {
+    return value;
```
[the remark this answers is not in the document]
ciaranschutte: we will always have a `value` or our type check will fail
but your point stands regarding the optional `field`

changing so if we get just a value, and it doesn't match any special display condition, value is simply returned

joneubank (approved):

PR #1828, opened 2021-02-18
Feat/#1819 dashboard UI

ciaranschutte:
**Description of changes**

- remove extended fields and processing status column
- remove extended fields from data query
- comment out stats we don't need (they may come back)

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


anncatton (approved):

components/pages/submission-system/program-dashboard/StatsBar/index.tsx
```diff
@@ -173,6 +173,7 @@ export default () => {
               </Statistic>
             )}
           </Col>
+          {/*
```
joneubank, line 176: comment block is way too big and is gonna be a pain until it gets restored... instead, could you move this section of code to a function that would render the stat block, then call that function here and comment out just the one line calling the function? Leave a comment as to why this is commented out.

joneubank (changes requested):

components/pages/submission-system/program-dashboard/StatsBar/index.tsx
ciaranschutte: @joneubank checked with Rosi. we don't need this ever again. ever. deleted.

joneubank: thanks!

joneubank (approved):

PR #1831, opened 2021-02-19
✨ clinical chart - proof of concept

samrichca:
**Description of changes**

Add a feature flagged, proof of concept version of clinical chart to the program dashboard, using mock data.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


global/utils/common.tsx
```diff
@@ -44,10 +44,10 @@ export const asEnum = (obj, { name = 'enum' } = {}) =>
     },
   });
 
-const standardDate = 'YYYY-MM-DD';
```
samrichca, line 47: when i updated date-fns this started throwing errors

package.json
```diff
@@ -57,7 +58,7 @@
     "babel-polyfill": "^6.26.0",
     "clsx": "^1.0.4",
     "color": "^3.1.2",
-    "date-fns": "^1.30.1",
+    "date-fns": "^2.17.0",
```
ciaranschutte, line 61: Did this break the other uses of `date-fns`?
'/global/utils/common.tsx'
'/components/pages/submission-system/program-dashboard/DonorDataSummary/DownloadButtons.tsx'

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
```diff
@@ -0,0 +1,284 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import React from 'react';
+import { differenceInDays, eachQuarterOfInterval, getQuarter, getYear, isAfter, isBefore, compareAsc } from 'date-fns';
+import * as u from './utils';
+import * as t from './types';
```
ciaranschutte, line 23: stylistically wondering if we should stick with explicitly importing like:
`import { differenceInDays, eachQuarterOfInterval, getQuarter, getYear, isAfter, isBefore, compareAsc } from 'date-fns';`

Are there benefits to using the `import * ` syntax?

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
```diff
@@ -0,0 +1,284 @@
...
+      <g
+        fill="none"
+        stroke={options.colors.axisBorder}
+        strokeWidth={options.strokeWidth}
+        >
+        {new Array(xTicksCount).fill(0).map((_, index) => {
+          const tickX = getX(index);
+          return (
+            <polyline
+              key={_}
```
ciaranschutte, line 187: is this just to shut up a react warning?

components/pages/submission-system/program-dashboard/LineChart/RangeControlBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+  isActive: boolean;
+  label: string;
+}) => {
+  return (
+    <button
+      aria-label={label}
+      title={label}
+      type="button"
+      onClick={handleClick}
+      style={{...styles.button, ...(isActive ? styles.buttonActive : {})}}
```
ciaranschutte, line 79: This is a different stylistically from most of the project which is using Emotion.
Is is possible to use it here?

something like:
```
<button css={css`
  background: transparent;
    border: 0 none;
    color: ${isActive? 'red' : #525767};
    cursor: 'pointer';
    display: inline-block;
    fontFamily: 'Work Sans, sans-serif';
    fontSize: 12px
    padding: 4;
    width: 25px;
   
`}...../>
```

ciaranschutte:
anything to gain from using d3?

components/pages/submission-system/program-dashboard/LineChart/RangeControlBar.tsx
samrichca: sure thing

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
joneubank: If we are only using part of the library, then using the explicit imports has the potential to reduce the size of the JS files needed to be downloaded on the UI. Therefore, explicit import from bigger libraries (ie. lodash) is prefered, and should be our standard for dependencies including internal dependencies from things like `utils` and `types`

If we do use the `import * as utils from './utils';` we should not be using single letter names.

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
joneubank: I'm curious where you got this as well. Since this is JS, what value do we expect the underscore to output in this case, seems like an intriguing solution.

.env.schema
```diff
@@ -39,3 +39,7 @@ TEST_GOOGLE_PASS= # user pass
 
 # For Storybook only
 STORYBOOK_GATEWAY_API_ROOT=http://localhost:9000
+
+# Feature flag for charts on program dashboard
+# (Completed Core Clinical Data & Molecular Data Summary)
+DASHBOARD_CHARTS_ENABLED=true
```
joneubank, line 45: Lets use the word FEATURE in all our feature flag variables: `FEATURE_DASHBOARD_CHARTS_ENABLED`

Unfortunately there are many other env variables used that are not in this schema file... we should update that in the future.

global/config.ts
```diff
@@ -43,6 +43,7 @@ export const getConfig = () => {
     FEATURE_LANDING_PAGE_STATS_ENABLED: publicConfig.FEATURE_LANDING_PAGE_STATS_ENABLED === 'true',
     FEATURE_ACCESS_FACET_ENABLED: publicConfig.FEATURE_ACCESS_FACET_ENABLED === 'true',
     MAINTENANCE_MODE_ON: publicConfig.MAINTENANCE_MODE_ON === 'true',
+    DASHBOARD_CHARTS_ENABLED: publicConfig.DASHBOARD_CHARTS_ENABLED === 'true',
```
joneubank, line 46: update to `FEATURE_DASHBOARD_CHARTS_ENABLED`

components/pages/submission-system/program-dashboard/CompletedClinicalData/index.tsx
```diff
@@ -36,23 +42,49 @@ const getStartedLink = (
...
+        Completed Core Clinical Data
+      </Typography>
+      <div
+        css={css`
+          height: 260px;
+          padding: 12px 0;
+          /* display: flex;
+          flex-direction: column;
+          justify-content: center;
+          align-items: center; */
```
joneubank, line 70: Remove instead of commenting.

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+const styles = {
+  axisLabel: {
+    fill: options.colors.text,
+    fontFamily: options.fontFamily,
+    fontSize: options.fontSize,
+  }
+};
+
+const LineChartEl = ({
```
joneubank, line 49: Reading `El` is confusing and it appears to not be standard in the platform components. Does it stand for 'element'? Let's find a name that reflects what this component is or at very least not use the abbreviated version (full words unless very obvious names please)

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+import React from 'react';
+import { differenceInDays, eachQuarterOfInterval, getQuarter, getYear, isAfter, isBefore, compareAsc } from 'date-fns';
+import * as u from './utils';
+import * as t from './types';
+
+const options = {
+  colors: {
+    // colours are changed slightly from theme.
+    // svg rendering made the colours darker/lighter.
```
joneubank, line 28: 👍 Let's make sure we get Kim to review the chart before we publish. Thanks for this comment to explain the change in values.

Maybe we should move these values to an external constants file for SVG chart element properties... don't need to now, but it would be useful for future charts types that need consistent styling.

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+  // setup chart dimensions
+  const padding = (options.fontSize + yAxisDigits) * 3;
+  const chartWidth = width - padding * 1.5;
+  const chartHeight = height - padding * 2;
+
+  // setup X axis
+  // X axis ticks, labels, and line/point positions
+  // are 1/2 tick distance from the left and right
+  const dataPoints = data.lines[0].points;
```
joneubank, line 77: Will there be issues if other lines have more/less points?

components/pages/submission-system/program-dashboard/LineChart/RangeControlBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import React from 'react';
+import * as u from './utils';
+
+const styles = {
```
joneubank, line 23: Take a look at how `styled` elements are made and used in other components, very similar to what you have here but can import theme properties. We may be able to include the svg colours in the theme then.

joneubank (changes requested):
Lots of amazing work here, just called out some simple things in comments (should've grouped them as a review but I messed up...) . Thanks Sam!

package.json
samrichca: it broke both but i only fixed one, good catch

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
samrichca: I did indeed use explicit imports from libraries for performance/build size reasons 

Updated utils/types to be explicit imports

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
samrichca: @ciaranschutte Yeah, not supposed to use index for keys in map functions because index could change

@joneubank the underscores are from an example I started from, basically means, "this value doesn't matter". I updated the code to remove them

samrichca:
> anything to gain from using d3?

In this case no, since there's no heavy animation/customization/etc and I expect to handle the math on the backend.

@joneubank shared a great article with me: https://medium.com/@PepsRyuu/why-i-no-longer-use-d3-js-b8288f306c9a

components/pages/submission-system/program-dashboard/LineChart/LineChartEl.tsx
samrichca: For this particular chart I want all lines to have the same X coordinates. Potentially in the future I'd have to account for more/less points or points being different between lines but in this case we're aggregating data into 7 buckets on the backend, so there should be 7 points, although some might have null Y coordinates (I'll check with Rosi)

samrichca: Ok checked in with her:

- All lines start at program create date
- If the user queries a date before the program create date, we'll correct it to show program create date to [present or whatever the end date was in the query]

joneubank: in this case, should we specify the number of points and their labels separate from the data?

joneubank: Just trying to avoid mixups with the data and the desired chart appearance, in particular since it is possible to have multiple lines that are missing data points at the start/end of the time series.

samrichca: > in this case, should we specify the number of points and their labels separate from the data?

Yeah, I was thinking of doing this:

```
{ 
  chartType: 'molecular',
  committedDonors: 2000,
  dates: [ 7 dates in unix epoch ],
  lines: [{
    label: 'molecular lines have names',
    data: [ 7 y axis values ],
  }]
}
```

In this case there won't be any missing values, but there could be in some other line chart we eventually make. I'm not sure how general purpose I should make this UI component or the API

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -71,7 +71,7 @@ import useClickAway from 'uikit/utils/useClickAway';
 import TooltipFacet from './TooltipFacet';
 import { getConfig } from 'global/config';
 import useAuthContext from 'global/hooks/useAuthContext';
-import { FilterOption } from 'uikit/dist/OptionsList';
+import { FilterOption } from 'uikit/OptionsList';
```
samrichca, line 74: this was throwing an error

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -414,7 +414,7 @@ export default () => {
     ).toString();
   };
 
-  const facetContianerLoadingStyle = css`
+  const facetContainerLoadingStyle = css`
```
samrichca, line 417: typo

ciaranschutte (approved):

components/pages/submission-system/program-dashboard/ClinicalChart/LineChart.tsx
```diff
@@ -0,0 +1,283 @@
...
+}: { 
+  data: DataObj;
+  hasQuarterLines: boolean;
+  height: number;
+  horizontalGuides: number;
+  precision: number;
+  width: number;
+}) => {
+  // setup Y axis
+  // TODO: maxY is this OR committed donors, whichever is more
```
joneubank, line 64: This is specific to the donor page usage of the line chart, not specific to Line Charts themselves. Ideally we would keep this as a generic LineChart component to be used in several places around the platform, and we can make a specific version that includes horizontal lines that show the Committed Donors just for the Donor Dashboard.

samrichca: agreed, I wanna get it working and then make it more general purpose

joneubank (approved):
Great start to building this chart.

As discussed, this is work in progress. We plan to refactor in the future to make the LineChart a generic chart component that is provided by the UI kit, but we are building in context for now to speed up development. Future work will be done to separate it and make a generic interface for chart properties such as data/axis and labels/toggles for grid lines, etc.

components/pages/submission-system/program-dashboard/DonorDataSummary/DownloadButtons.tsx
```diff
@@ -106,7 +106,7 @@ const DownloadButtons = () => {
       .then(checkResponseIs200)
       .then(res => res.blob())
       .then(blob => {
-        const now = formatDate(Date.now(), 'YYYYMMDD');
+        const now = formatDate(Date.now(), 'yyyyMMdd');
```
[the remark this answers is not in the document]
samrichca: that's a good idea

global/utils/common.tsx
```diff
@@ -44,12 +44,19 @@ export const asEnum = (obj, { name = 'enum' } = {}) =>
...
-export const displayDate = (date) => formatDate(date, standardDate);
+const dateFormat = 'yyyy-MM-d';
```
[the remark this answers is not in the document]
samrichca: good catch, fixed

Issue #1851, opened 2021-03-08
🐛 "Non Nullable" Error causes dashboard page to crash. 


rosibaj:
## Steps To Reproduce

1. Go to  https://platform-ui.qa.argo.cancercollaboratory.org/submission/program/DOG-CA/dashboard
2. See the page crash, with these errors: 
![image](https://user-images.githubusercontent.com/15824252/110387170-3a029c80-802f-11eb-8dd5-f5c08773d370.png)

## Expected behavior
The UI should not crash like this when it gets errors.  Whatever CARD is having an issue maybe shouldnt load, or should load with an error message but the page should load. 

![image](https://user-images.githubusercontent.com/15824252/110666420-53265d00-8197-11eb-99aa-57be3d956ac1.png)

https://zpl.io/2jNPNXO

rosibaj:
checked on qa PROGRAM DOG-CA. which has an index issue. 

The UI is correctly displaying the error: 
![image](https://user-images.githubusercontent.com/15824252/111798714-b6ec0c80-88a0-11eb-9163-f78a9ee8fdae.png)

L:GTM thanks @samrichca !

Issue #1844, opened 2021-03-03
🐛 Fix the width of the edit user modal


kcullion:
## Describe the bug
The `edit user` modal is using a smaller width. Please make it the same as the `add user` modal width

## Steps To Reproduce
1. Go to a `Manage Program` for DASH-CA, for example
2. Click on the pencil icon for Rosi's data submitter user
3. The modal is very squished:
<img width="847" alt="2021-03-02_16-34-41" src="https://user-images.githubusercontent.com/18424138/109826408-c3b60280-7c08-11eb-8e9d-682625693ab5.png">

## Expected behaviour
Please use the same width as the add users modal:
![image](https://user-images.githubusercontent.com/18424138/109826495-d7f9ff80-7c08-11eb-9575-f0098ceccb4f.png)

ciaranschutte:
I had trouble recreating this on `develop`

rosibaj:
@ciaranschutte i still see this on the qa env: 
![image](https://user-images.githubusercontent.com/15824252/110387589-d036c280-802f-11eb-939e-3141848b9bfc.png)

ciaranschutte:
I see... I've left some comments on Sam's PR

rosibaj:
lgtm on QA thanks @samrichca and @ciaranschutte !

rosibaj:
verified on production with OCCAMS-GB user

PR #1874, opened 2021-03-18
rc/1.90.0

samrichca:
- make Edit Users modal the same width as Add Users modal https://github.com/icgc-argo/platform-ui/issues/1844
- fix non-nullable error in program dashboard https://github.com/icgc-argo/platform-ui/issues/1851


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "argo-platform-ui",
-  "version": "1.88.2",
```
samrichca, line 3: this is `1.89.0` on master, not sure why it's `1.88.2` here

joneubank (approved):

PR #1877, opened 2021-03-29
Add DonorAggregationSummary table to DCC Dashboard

joneubank:
**Description of changes**
Adding a way for a DCC Admin to see when the Donor Aggregation Index for each program was last updated and trigger a new SYNC for that index if needed. The table includes a summary of last update, number of donors and number of files on the donor aggregation index.

This uses the new `/donor-aggregation/sync` endpoint added to the gateway here: https://github.com/icgc-argo/platform-api/pull/382

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

**Details**
_Data Fetching and Loading_
The card will get all Programs in one request then passes these programShortNames to the table. The table fetches the donor summary stats data for each row from the gateway (could be simplified with modifications to the gateway allowing a bulk fetch). The table will load in each row of data as it comes but will remain in the "loading" state until all rows are fetched. This is done to prevent interaction with the SYNC buttons before the rows are loaded since the button state will be reset when each new row of data is returned.

This can be improved by managing the data fetching and button state in the Table level, but in practice shouldn't be an issue. Can refactor to this if the Program list gets long in prod like it is in QA.

_Appearance_
Loaded State:
Screen Shot 2021-03-29 at 1 06 35 PM

SYNC success state:
Screen Shot 2021-03-29 at 1 35 07 PM

SYNC error state:
Screen Shot 2021-03-29 at 1 27 23 PM

- [ ] Matches design:

  - font size, weight, colour


package.json
```diff
@@ -114,12 +107,20 @@
     "@storybook/addon-viewport": "^6.0.22",
     "@storybook/addons": "^6.0.22",
     "@storybook/react": "^6.0.22",
+    "@types/enzyme-adapter-react-16": "^1.0.5",
     "@types/file-saver": "^2.0.1",
+    "@types/graphql": "^14.5.0",
+    "@types/isomorphic-fetch": "0.0.35",
```
joneubank, line 113: This one is new. Rest were just moved to dev-dependencies.

uikit/Typography/stories.tsx
```diff
@@ -33,7 +33,11 @@ const TypographyStories = storiesOf(`${__dirname}`, module)
         Object.keys(defaultTheme.typography) as Array<TypographyVariant>,
         'hero',
       ),
-      component: select('tag', [null, 'h1', 'h2', 'h3', 'h4', 'h5', 'div', 'span', 'p'], null),
+      component: select(
+        'component',
```
joneubank, line 37: Changed from `tag` to `component` to match the prop name in the Typography component.

joneubank:
Since taking those screenshots I updated the link stylign to match argo. Everything else is the same, just the typography and text color of the links now matches the site.

components/pages/submission-system/dcc/DonorAggregationSummary/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+// GQL Data Fetching
+import { useQuery } from '@apollo/react-hooks';
+import SIDE_MENU_PROGRAM_LIST from 'components/pages/submission-system/SIDE_MENU_PROGRAM_LIST.gql';
+
+const Card = styled(Container)`
+  padding: 16px;
+`;
+
+const DonorAggregationSummary = () => {
+  // const { loading, data } = useDonorAggregationIndexData();
```
ciaranschutte, line 35: @joneubank leftover comment?

components/pages/submission-system/dcc/DonorAggregationSummary/table-cell-components/SyncIndexButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+import Typography from 'uikit/Typography';
+
+import useAuthContext from 'global/hooks/useAuthContext';
+import { getConfig } from 'global/config';
+const { GATEWAY_API_ROOT } = getConfig();
+import { DONOR_AGGREGATOR_SYNC_PROGRAM } from 'global/constants/gatewayApiPaths';
+
+const SyncIndexButton = ({ program }: { program: string }) => {
+  const theme = useTheme();
+  const [requestResult, setRequestResult] = useState<'SUCCESS' | 'ERROR' | null>(null);
```
ciaranschutte, line 35: worth using an enum seen as we're using typescript or even just a js constant for SUCCESS and ERROR?

components/pages/submission-system/dcc/DonorAggregationSummary/table-cell-components/ProgramDashboardLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import Link from 'next/link';
+import A from 'uikit/Link';
+import { PROGRAM_SHORT_NAME_PATH, PROGRAM_DASHBOARD_PATH } from 'global/constants/pages';
+const ProgramDashboardLink = ({ program }: { program: string }) => {
+  return (
+    <Link
+      as={PROGRAM_DASHBOARD_PATH.replace(PROGRAM_SHORT_NAME_PATH, program)}
+      href={PROGRAM_DASHBOARD_PATH}
```
ciaranschutte, line 27: we have this component everywhere loosely instead of uikit.
no need to fix in this PR just thinking out loud

components/pages/submission-system/dcc/DonorAggregationSummary/table-cell-components/SyncIndexButton.tsx
joneubank: Since this is only used locally and we only need these 3 states just declaring them in here should be sufficient for type checking its usage. Not exporting the type or the hook.

components/pages/submission-system/dcc/DonorAggregationSummary/DonorAggregationIndexTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+    shortName,
+    donors: data?.programDonorSummaryStats.registeredDonorsCount,
+    files: data?.programDonorSummaryStats.allFilesCount,
+    lastUpdate: data?.programDonorSummaryStats.lastUpdate,
+  }));
+
+  const someQueriesLoading = queries.some((query) => query.query.loading);
+
+  const containerRef = React.createRef<HTMLDivElement>();
+  return (
```
ciaranschutte, line 94: can't recall off hand if `Table` component displays a `No Data` component for an empty array.
If it doesn't, may be worth just returning an `No Data` component immediately if we don't have programs

components/pages/submission-system/dcc/DonorAggregationSummary/table-cell-components/ProgramDashboardLink.tsx
joneubank: Same with adding `padding: 16px` to Container and calling it `Card` . Can do a refactor where we pull out these kind of usages.

For `<Link><A></A></Link>` we could add a property to `uikit/Link` for `internal:boolean` that would add the `nextjs/Link` wrapper.

components/pages/submission-system/dcc/DonorAggregationSummary/DonorAggregationIndexTable.tsx
joneubank: Seen No Data state in testing, so it seems the `Table` handles this just fine.

ciaranschutte:
looks like CI is failing because the added `lastUpdate` field isn't deployed so our gql query is failing a test

ciaranschutte (approved):

PR #1879, opened 2021-03-31
Feat/#1855 file entity page

ciaranschutte:
**Description of changes**

Adds file entity page to platform ui
- behind feature flag: FEATURE_FILE_ENTITY_ENABLED
- does page level gql query check to make sure user is navigating to valid file entity

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour


pages/file/[fileId].tsx
```diff
@@ -0,0 +1,70 @@
...
+      const err = new Error('Page Not Found') as Error & { statusCode?: number };
+      err[ERROR_STATUS_KEY] = 404;
+      throw err;
+    }
+  },
+})((props) => {
+  const { fileId } = usePageQuery<{ fileId: string }>();
+  const filters = sqonBuilder.has('object_id', fileId).build();
+
+  // small query to ensure the fileId is valid
```
Comment (ciaranschutte) [inline pages/file/[fileId].tsx:46]: check if the fileId in the url is valid otherwise show an error page

components/pages/file-entity/FileCardsLayout.tsx
```diff
@@ -38,11 +38,13 @@ const FileCardsLayout: React.ComponentType<{
           <AssociatedDonors donors={fileData.donorRecords} />
         </PaddedColumn>
       </PaddedRow>
+      {/*
```
ciaranschutte, line 41: comment out complete but inactive section for now

components/pages/file-entity/useEntityData.tsx
```diff
@@ -0,0 +1,92 @@
...
+          matchedNormalSampleId: sample.matched_normal_submitter_sample_id,
+        },
+      };
+    });
+
+    const entityData: FileEntityData = {
+      summary,
+      dataAnalysis,
+      donorRecords,
+      // fileRecords: dummyFileRecords,
```
ciaranschutte, line 85: comment out complete but inactive section for now

pages/file/[fileId].tsx
```diff
@@ -0,0 +1,70 @@
...
+    },
+  });
+  const isValidEntity = !!get(data, 'file.hits.total', false);
+
+  const { setLoading: setLoaderShown, isLoading: isLoaderShown } = useGlobalLoadingState();
+
+  if (!loading && !isValidEntity) {
+    setLoaderShown(false);
+    const err = new Error('Page Not Found') as Error & { statusCode?: number };
+    err[ERROR_STATUS_KEY] = 404;
```
Comment (ciaranschutte) [inline pages/file/[fileId].tsx:61]: returning <Error> to show an error page
usually we are throwing an error but this leaves a blank page for user

pages/file/[fileId].tsx
```diff
@@ -0,0 +1,70 @@
...
+  }>(VALID_ENTITY_CHECK, {
+    variables: {
+      filters,
+    },
+  });
+  const isValidEntity = !!get(data, 'file.hits.total', false);
+
+  const { setLoading: setLoaderShown, isLoading: isLoaderShown } = useGlobalLoadingState();
+
+  if (!loading && !isValidEntity) {
```
Comment (ciaranschutte) [inline pages/file/[fileId].tsx:58]: @anncatton  do I have to check if a user is loggingOut too?
I saw similar logic on the submission pages

components/pages/file-entity/index.stories.tsx
```diff
@@ -19,10 +19,64 @@
...
+import clsx from 'clsx';
+import Head from '../head';
+import DnaLoader from 'uikit/DnaLoader';
+import Footer from 'uikit/Footer';
+import { PageContainer, PageBody, PageContent, ContentHeader, ContentBody } from 'uikit/PageLayout';
+import FileCardsLayout from './FileCardsLayout';
+import { FileTitleBar } from './FileTitleBar';
+import NavBar from '../../NavBar';
 
 const FileRepositoryTableStories = storiesOf(`${__dirname}`, module).add('Basic', () => {
```
ciaranschutte, line 38: Data fetching happens in the component.
There's no point changing the component to suit the storybook usage so I've pretty much copied the page in here with the dummy data.

pages/file/[fileId].tsx
Comment (anncatton) [reply pages/file/[fileId].tsx:58]: `isLoggingOut` was added to prevent the `useProgramCheckEffect` getting called and throwing an error when logout was initiated but the user was still on a page that was trying to react to the token state...has any of the logic surrounding that changed? It's been awhile since that got put in

Comment (ciaranschutte) [reply pages/file/[fileId].tsx:58]: No logic has changed around that in this PR.
We've building fairly common patterns in those page component so just making sure I have everything covered.

components/pages/file-entity/FileCardsLayout.tsx
joneubank: Why is this inactive and when will it be activated? Could you leave a code comment?

components/pages/file-entity/useEntityData.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  if (loading) {
+    return { programShortName: null, data: null, loading: true };
+  } else {
+    if (!data) {
+      return { programShortName: null, data: null, loading: false };
+    }
+
+    const entity = get(data, 'file.hits.edges[0].node');
```
joneubank, line 34: This could come back as undefined, and then all of the property references below will throw errors. We need to handle that missing value case.

components/pages/file-entity/useEntityData.tsx
joneubank: code comment to explain, maybe a TODO

components/pages/file-entity/useEntityData.tsx
```diff
@@ -0,0 +1,92 @@
+import { useQuery } from '@apollo/react-hooks';
```
joneubank, line 1: Copyright notice added to files you modify? Plzkthx?

components/pages/file-entity/FileTitleBar.tsx
```diff
@@ -4,13 +4,20 @@ import TitleBar from 'uikit/TitleBar';
...
+import sqonBuilder from 'sqon-builder';
 
 export const FileTitleBar: React.ComponentType<{
   programShortName: string;
   fileId: string;
   isUserLoggedIn: boolean;
 }> = ({ programShortName, fileId, isUserLoggedIn }) => {
   const theme = useTheme();
+  const { GATEWAY_API_ROOT } = getConfig();
+  const filter = sqonBuilder.has(FileCentricDocumentField['object_id'], fileId).build();
```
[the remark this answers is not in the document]
anncatton: sqon builder all over the file repository now :)

pages/file/[fileId].tsx
Comment (anncatton) [reply pages/file/[fileId].tsx:58]: I'm thinking, if a file entity can have controlled access, it would be good to do the `isLoggingOut` check, so you don't get an unnecessary 403, to prevent the page from trying to fetch a file after the token has been removed on logout

components/pages/file-entity/FileCardsLayout.tsx
ciaranschutte: @joneubank might need RDPC api for this, might not. Spec not ready yet. 
Comment as vague as this may be pointless?

ciaranschutte: Just removing the unused code.
It should be clear from ticket comments that we have that last bit of UI already made.

It's the same flow has this ticket is.
Taking UI already made and adding it to a page structure.

Therefore a scattering of TODOs is not needed,it's just cruft

ciaranschutte:
@joneubank addressed your feedback. 
I just removed unused code instead of sprinkling TODOs.
We are aware the component is there for future use, I think that's sufficient.

components/pages/file-entity/useEntityData.tsx
```diff
@@ -47,12 +47,12 @@ const useEntityData = ({ fileId }: { fileId: string }): EntityData => {
   if (loading) {
     return { programShortName: null, data: null, loading: true };
   } else {
-    if (!data) {
+    const entity = get(data, 'file.hits.edges[0].node', null);
+
+    if (!entity) {
```
joneubank, line 52: 👍

joneubank (approved):
Thanks Ciaran!

PR #1881, opened 2021-04-05
Clinical Timeline updates

ciaranschutte:
**Description of changes**

Clinical Timeline: Treatment tab update 
visible in storybook in: http://localhost:6006/?path=/story/components-pages-submission-system-donor-clinicaltimeline--basic

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour

### DIFF
--- a/components/pages/submission-system/donor/ClinicalTimeline/index.tsx
+++ b/components/pages/submission-system/donor/ClinicalTimeline/index.tsx
@@ -20,20 +20,19 @@
 import React from 'react';
 import Container from 'uikit/Container';
 import { css } from 'uikit';
-import Header from './header';
-import Timeline from './timeline';
+import Header from './Header';
+import Timeline from './Timeline';
 import { EntityType } from './types';
 import Typography from 'uikit/Typography';
 import SimpleTable from 'uikit/Table/SimpleTable';
 import get from 'lodash/get';
-import Samples from './samples';
+import Samples from './Samples';
 import { Row, Col } from 'react-grid-system';
 import { useTheme } from 'uikit/ThemeProvider';
 import ContentPlaceholder from 'uikit/ContentPlaceholder';
-import Table from 'uikit/Table';
-import { createRef } from 'react';
 import chunk from 'lodash/chunk';
 import { isEmpty } from 'lodash';
+import Treatment, { ITreatment } from './Treatment';
 
 export const ENTITY_DISPLAY = Object.freeze({
   primary_diagnosis: {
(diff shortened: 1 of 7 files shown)


components/pages/submission-system/donor/ClinicalTimeline/index.tsx
```diff
@@ -121,25 +188,50 @@ const ClinicalTimeline = ({ data }) => {
...
+                    flex-direction: column;
                   `}
                 >
-                  <SimpleTable data={activeEntityData} />
+                  {
+                    <Row>
+                      <Col>
+                        <SimpleTable data={tableFormat(dataCol0)} />
+                      </Col>
+                      {selectedSamples.length === 0 && !isEmpty(selectedData) && (
```
ciaranschutte, line 205: There is a variable amount of columns for the clinical timeline
If we have `data` and `samples` we show all the data in one col, and the samples in the other
If we only have `data` we have two columns

The `treatment` sections comes after

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
```diff
@@ -47,24 +50,90 @@ export const ENTITY_DISPLAY = Object.freeze({
...
+          showPagination={false}
+          sortable={false}
+        />
+      </div>
+    </div>
+  );
+};
+
+// chunk object to show as 2 cols
+const chunkObject = (data, size: number): any[][] =>
```
joneubank, line 94: This method is not doing what the comment suggests, it is doing two things:
1. Convert an object into an array of its entries
2. Applying _.chunk with the size value provided in the arguments.

Because size is provided by the caller, this implementation will return a variable number of columns (not guaranteed to be 2).

I would like to see two things changed:
1. Rename to be more specific than chunk, perhaps: `splitIntoColumns`
1. `data` needs types declared so we now it is accepting only objects not string/number etc. should be `{[key:string]:any}` I think?
1. Instead of accepting the size of the column as an input, instead take the number of columns. You can use that to calculate size of each chunk by dividing `Objects.entries(data).length / numberOfColumns` and use this in the chunk operation

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
joneubank: I don't like how the `selectedSamples.length > 0` check is repeated in separate places (here and where you call the chunk method)... feel like thats asking for future bugs.

I would simplify this section by handling all this logic inside this section, so it would read like:

```
<row>
if(selectedSamples.length > 0) { 
  1. splitInoColumns(data, 1) // see my comment on the 'chunk' method
  2. <col> data </col><col>samples</col>
} else {
  1. splitInoColumns(data, 2)
  2. <col>data column 1</col> <col>data column 2</col>
}
</row>
```
Since this is now all self contained, you could move the entire row into its own method, something like `renderSelectedDataRow(selectedData, selectedSamples)`

joneubank (changes requested):
Functionally good, just ideas on re-ogranizing/simplifying how the code reads.

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
ciaranschutte: sleepy comment. great suggestions thank you

ciaranschutte:
@joneubank 
implemented your great suggestions. 
added some additional commenting.
cleaned up some naming and imports/file structure

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
```diff
@@ -130,7 +114,7 @@ const ClinicalTimeline = ({ data }) => {
   const selectedTreatments: ITreatment[] = get(selectedClinical, 'treatments', []);
   const selectedData = get(selectedClinical, 'data', {});
 
-  const [dataCol0 = [], dataCol1 = []] = chunkObject(
+  const [dataCol0 = [], dataCol1 = []] = splitIntoColumns(
```
joneubank, line 117: these consts are never used (I think, checked pretty quick...) . Can be removed.

joneubank (approved):

components/pages/submission-system/donor/ClinicalTimeline/index.tsx
ciaranschutte: nice catch

PR #1893, opened 2021-04-19
remove additional token state

ciaranschutte:
**Description of changes**

Auth Provider had both token state and a token prop.
This causes a conflict when a new token comes from `_app`
Removed this extra state.

@anncatton I believe you worked on this section last, so please take a look in case I've overlooked something.
In particular regarding refreshed JWTs vs initial.
I've tested it locally and it seems to work. Logging/logging out is always combined with routing to a new page so `_app` re-renders and passes down the most recent `egoJwt`

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
- [ ] Feature is minimally responsive
- [x] Connected ticket to PR

### DIFF
--- a/global/hooks/useAuthContext.tsx
+++ b/global/hooks/useAuthContext.tsx
@@ -28,7 +28,7 @@ import urlJoin from 'url-join';
 import refreshJwt from 'global/utils/refreshJwt';
 
 type T_AuthContext = {
-  token?: string;
+  egoJwt?: string;
   logOut: (config?: { toRoot?: boolean }) => void;
   updateToken: () => Promise<string | void>;
   data: ReturnType<typeof decodeToken> | null;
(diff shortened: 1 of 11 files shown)


global/hooks/useAuthContext.tsx
```diff
@@ -62,21 +61,22 @@ export function AuthProvider({
     `/api/oauth/update-ego-token?client_id=${EGO_CLIENT_ID}`,
   );
 
-  const [token, setTokenState] = React.useState<string>(egoJwt);
-  const [permissions, setPermissions] = React.useState<Array<string>>(initialPermissions);
+  const token = egoJwt;
```
joneubank, line 64: why do this?

joneubank (commented):
It looks like we only need to use the `egoJwt` variable now, no need to assign to `token`... any reason you didn't convert all references to `egoJwt`?

ciaranschutte:
@joneubank updated usage of `token` to `egoJwt` across the project

joneubank (approved):

Issue #2009, opened 2021-07-29
Connect alignment column to filters


rosibaj:
Reference for storybook for the UI component library: https://argo-ui-storybook.netlify.app/

https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=icgcargotech&title=Program+Dashboard+Page+Specs#ProgramDashboardPageSpecs-4.1TableDataFields,Filters,&DataSources

- Add the `Alignment` filter to the UI 
![image](https://user-images.githubusercontent.com/15824252/127512652-d7430808-f65f-445e-9dec-16518ec3c977.png)

rosibaj:
confirmed in production release sept 02, 2021

Issue #2008, opened 2021-07-29
Connect Raw Reads + Samples columns to filters 


rosibaj:
Reference for storybook for the UI component library: https://argo-ui-storybook.netlify.app/

https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=icgcargotech&title=Program+Dashboard+Page+Specs#ProgramDashboardPageSpecs-4.1TableDataFields,Filters,&DataSources

- Add the `Raw Reads` and `Samples` filter to the UI 
![image](https://user-images.githubusercontent.com/15824252/127512456-7bb313d5-d985-4f6d-b522-0512c5627961.png)

Issue #2001, opened 2021-07-15
Feature Request: Connect Filters to Dashboard - Core Clinical 


rosibaj:
Reference for storybook for the UI component library: https://argo-ui-storybook.netlify.app/

https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=icgcargotech&title=Program+Dashboard+Page+Specs#ProgramDashboardPageSpecs-4.1TableDataFields,Filters,&DataSources

As discussed on Slack...The following outlines the general filter behaviour for the columns (for testing too)

- [x] The filter icon when hovering changes from grey to the blue. Nice to have: a tooltip that says "Filter by column name" when hovering over the filter icon
- [x] The user clicks on the filter icon to open the filter options. It closes when they click on that filter icon OR clicks on another filter icon (two filters would not be open at the same time) OR clicks anywhere else on the screen besides the filter option dropdown
- [x] Clicking anywhere else on the table header cell will sort the column as it's working now
- [x] When the filter is open, the active state would be the whole header cell gets shaded grey and the filter icon is blue.

supported by: https://github.com/icgc-argo/platform-api/issues/416

rosibaj:
confirmed in production release sept 02, 2021

PR #2025, opened 2021-08-20
✨ Add List-style Dashboard Dropdown Filters (#2001 #2008 #2009)

mistryrn:
**Description of changes**

**TL;DR: creates `ListFilter` inside `DropdownPanel`, a fully accessible checkbox list filter panel, and applies them to the headers of the "Core Completion", "Samples", "Raw Reads", and "Alignment Status" columns in the Dashboard table. Also includes some refactoring and bug fixes for the existing `TextInputFilter` component added previously.**

The biggest addition in this PR is the new `ListFilter` component within `DropdownPanel` of UIKit. It satisfies the design and behaviour requirements of https://github.com/icgc-argo/platform-ui/issues/2001 , which I felt are different enough from the existing `OptionsList` component in UIKit to warrant its own component. This is fully keyboard accessible, and has props for providing specific labels for screen readers if needed.

List filters have been added to the headers of the following columns of the Dashboard table: "Core Completion" (#2001), "Samples" (#2008), "Raw Reads" (#2008), and "Alignment Status" (#2009)

Due to the similarities between the `TextInputFilter` panel and the `ListFilter` panel, a great deal of refactoring was done to try to minimize any code duplication.

Storybook has been updated with examples of each type of dropdown filter, as well as a new generic "Basic" version of the `DropdownPanel` component.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
  - [x] Modified Existing components and updates stories
- [ ] Connected ticket to PR

### DIFF
--- a/components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
+++ b/components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
@@ -459,7 +459,29 @@ export default ({
         },
 
         {
-          Header: 'Samples',
+          Header: (
+            <ListFilterHeader
+              header={'Samples'}
+              panelLegend={'Sample Registration Status'}
+              onFilter={(options) =>
+                updateFilter(
+                  'registeredSamplePairs',
+                  options.filter((option) => option.isChecked).map((option) => option.key),
+                )
+              }
+              filterOptions={[
+                {
+                  key: 'VALID',
+                  value: 'Valid',
+                },
+                {
+                  key: 'INVALID',
+                  value: 'Invalid',
+                },
+              ]}
+              activeFilters={getFilterValue('registeredSamplePairs')}
+            />
+          ),
           id: REGISTERD_SAMPLE_COLUMN_ID,
           Cell: ({ original }) => (
             <DesignationCell
@@ -478,7 +500,29 @@ export default ({
       },
       columns: [
         {
-          Header: 'Raw Reads',
+          Header: (
+            <ListFilterHeader
+              header={'Raw Reads'}
+              panelLegend={'Raw Reads Status'}
+              onFilter={(options) =>
+                updateFilter(
+                  'rawReads',
+                  options.filter((option) => option.isChecked).map((option) => option.key),
+                )
+              }
+              filterOptions={[
+                {
+                  key: 'VALID',
+                  value: 'Valid',
+                },
+                {
+                  key: 'INVALID',
+                  value: 'Invalid',
+                },
+              ]}
+              activeFilters={getFilterValue('rawReads')}
+            />
+          ),
           id: RAW_READS_COLUMN_ID,
           Cell: ({ original }) => (
             <DesignationCell
--- a/components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
+++ b/components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
@@ -31,7 +31,7 @@ import Table, { TableColumnConfig } from 'uikit/Table';
 
 import { displayDate } from 'global/utils/common';
 import Icon from 'uikit/Icon';
-import DropdownPanel, { TextInputFilter } from 'uikit/DropdownPanel';
+import DropdownPanel, { FilterOption, ListFilter, TextInputFilter } from 'uikit/DropdownPanel';
 import {
   DataTableStarIcon as StarIcon,
   TableInfoHeaderContainer,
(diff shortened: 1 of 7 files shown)


components/pages/submission-system/program-dashboard/DonorDataSummary/DonorSummaryTable.tsx
```diff
@@ -232,50 +301,83 @@ export default ({
...
+    onFilter = () => {},
+  }: {
+    header: string;
+    panelLegend?: string;
+    filterOptions?: Array<FilterOption>;
+    activeFilters?: Array<string>;
+    onFilter?: (filters?: Array<FilterOption>) => void;
+  }) => {
+    const [open, setOpen] = useState(false);
+    const options = React.useMemo(
```
mistryrn, line 342: Generating the options using `useMemo` to avoid re-renders of the child components.

In a future ticket, pending an API update, this will also be used to add the `doc_count` to each option.

uikit/DropdownPanel/index.tsx
```diff
@@ -86,6 +132,58 @@ export const DropdownPanelButtonSection = styled('div')`
   justify-content: space-between;
 `;
 
+export const FilterWrapper = ({
```
mistryrn, line 135: Refactoring (pt 4):
- `FilterWrapper` encapsulates the repeated components of the `DropdownPanel` between the text and list filters: form, fieldset, legend, and cancel/confirm buttons

uikit/DropdownPanel/index.tsx
```diff
@@ -104,13 +202,13 @@ export const TextInputFilter = ({
   confirmLabel?: string;
   inputLabel?: string;
   inputPlaceholder?: string;
-  onCancelClick?: (any?) => void;
-  onConfirmClick?: (any?) => void;
-  onInputChange?: (any?) => void;
+  onCancelClick?: (event?: any) => void;
```
mistryrn, line 205: 😅 missed the prop types here previously 🙈

uikit/DropdownPanel/index.tsx
```diff
@@ -121,56 +219,250 @@ export const TextInputFilter = ({
...
+      onCancelClick={(e) => {
+        e.preventDefault();
+        setOpen(false);
+        onCancelClick();
+      }}
+      onConfirmClick={(e) => {
+        e.preventDefault();
+        onConfirmClick(text?.trim());
+      }}
+      confirmDisabled={!text.length && !initialValue?.length}
```
mistryrn, line 235: Was able to fix an annoying behaviour here. Previously, the "Apply" filter button was disabled whenever the text input was empty. But that prevented you from clearing the filter, so I had it automatically clear inside the onChange whenever the text length went down to zero.

This was really annoying, as whenever you deleted all your input text the panel would immediately close because the filter was applied and the table would re-render.

Now, instead of all that, we just check to see if there is a filter currently applied using the `initialValue` prop. If there is, we are able to use the "Apply" button when there is no text in the input to clear if/when we want to.

uikit/DropdownPanel/index.tsx
```diff
@@ -121,56 +219,250 @@ export const TextInputFilter = ({
...
+          }}
+          ref={_inputRef}
+          showClear={true}
+        />
+      </DropdownPanelInputSection>
+    </FilterWrapper>
+  );
+};
+
+let listFilterInstances = 0;
```
mistryrn, line 257: Real juicy part of the PR is here (the `ListFilter` component).

Quick explainer for the "listFilterInstances" (there's a comment further down where it gets used but just in case):
- it's likely that we'll end up with multiple instances of these per page, so to ensure uniqueness among IDs for the options (but to still keep them predictable), we keep track of the count of instances and prefix the `instanceNum` to each option

uikit/DropdownPanel/index.tsx
```diff
@@ -121,56 +219,250 @@ export const TextInputFilter = ({
...
+    setOptions(options.map((option) => ({ ...option, isChecked: true })));
+    onSelectAllOptions();
+  };
+
+  const clear = () => {
+    setOptions(options.map((option) => ({ ...option, isChecked: false })));
+    onClear();
+  };
+
+  const focusNext = (i) => {
```
mistryrn, line 312: Keyboard a11y stuff specific to the checkbox lists from here to line 349:
- focus on the first checkbox when opening the panel
- can use the up/down arrows to navigate the list in addition to tab
- `home` goes to the first item, `end` goes to the last item
- can "loop" through the list (going "up" from the first takes you to the last, going "down" from the last takes you to the first) for ease of navigation through long lists

mistryrn (commented):
Adding some explanations of the refactors and some decisions made with the new components 👇

uikit/DropdownPanel/index.tsx
mistryrn: Bug ticket for the issue this fixes: #2027

PR #2053, opened 2021-09-14
add name to input in multiselect

samrichca:
**Description of changes**

- add name attribute to multiselect text input in the hopes it'll fire a change event on autocomplete in daco-ui
- for **single** multi-selects, add a change event that is exposed/bubbles to outer components.
- platform-ui didn't have issues in my testing ('create program' page in the UI & 'join program' in storybook)

re: https://github.com/icgc-argo/dac-ui/issues/211

### DIFF
--- a/uikit/form/MultiSelect/index.tsx
+++ b/uikit/form/MultiSelect/index.tsx
@@ -264,9 +264,9 @@ const MultiSelect = ({
   id = '',
   inputProps,
   name = '',
-  onBlur = (e) => {},
+  onBlur = (e) => { },
   onChange,
-  onFocus = (e) => {},
+  onFocus = (e) => { },
   placeholder,
   single = false,
   size = INPUT_SIZES.LG,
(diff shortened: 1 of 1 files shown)


caravinci (approved):
LGTM 👍

uikit/form/MultiSelect/index.tsx
```diff
@@ -334,7 +334,9 @@ const MultiSelect = ({
   };
 
   const handleInputChange = (event) => {
-    setSearchString(event.target.value);
+    const newValue = event.target.value;
+    setSearchString(newValue);
+    single && onChange(createCustomEvent(event, [newValue]), null);
```
caravinci, line 339: If this is a single selection, there's no reason why the value should be an array.
If an array output is needed at integration, it can be added on the receiving end (e.g. input handler).

```suggestion
    single && onChange(createCustomEvent(event, newValue), null);
```

samrichca: discussed on slack: this likely has to be an array, see `handleNewItemClick` above

caravinci: Confirmed. The pattern is not ideal (`single` vs `multiple` should produce clearly different results), but improving it would require additional work in Platform, etc.

i.e. to be improved upon at a later point.

PR #2065, opened 2021-09-21
update disabled colors in uikit

samrichca:
re: https://github.com/icgc-argo/dac-ui/issues/179

tested with storybook & setting fields on "create program" page to disabled

- make text inputs, checkboxes, radio buttons, select, and multiselect have grey text when disabled
- change textarea borders in disabled state

i refrained from reorganizing the colors, but i added a note in our uikit wiki page about organizing colors better.


anncatton (approved):
Oh my yes look at all those variations on a (disabled) theme...
thanks for making the wiki note @samrichca !

Issue #2148, opened 2021-12-21
File Entity: Add legend and private access tag


kcullion:
## Goal
Finish off the header of the file entity page.

## Logged in or logged out
- [x] Add the legend component to the page, built in: https://github.com/icgc-argo/uikit/issues/25 
![image](https://user-images.githubusercontent.com/18424138/146836537-dff77632-97da-479e-9767-090a564a9263.png)

## When logged in, as a program member/dcc admin
Zeplin: https://zpl.io/2E3g04n

- [x] add a tag to the file entity page, when the file has private access. 

So add a tag for a private file entity page that denotes the access tier:

- My Program Access (eg. https://platform-ui.qa.argo.cancercollaboratory.org/file/00095e36-49f6-520d-bb3a-a0ae8c31eb7e)
- Full Member Access (eg. https://platform-ui.qa.argo.cancercollaboratory.org/file/0e06fc30-651f-5144-91d9-aecfdafc8540)
- Associate Member Access (eg. collaboratory.org/file/001e0e50-85c3-52c5-b38c-67afca3aa044)

(the "Released to the Public" tier doesn't need a tag)
![image](https://user-images.githubusercontent.com/18424138/146967379-cd34eb95-d1bb-4290-b5fc-d3d4d8f156db.png)
![image](https://user-images.githubusercontent.com/18424138/146967408-e12eeaf8-c5e5-49b5-b66f-0d5089538dd6.png)

kcullion:
had a look in QA: the legend component has been added to the page and the access tier tags are displaying correctly. Nice job @demariadaniel !

kcullion:
working as expected in staging

PR #2152, opened 2022-01-07
💄 #2148 File Entity Private Access Tag

demariadaniel:
**Description of changes**

Adds a tag to the file entity page when the file belongs to a private access tier

**Type of Change**

- [ ] Bug
- [X] Styling
- [ ] New Feature

- Design (select one):
  - [X] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
  - [X] No new UIKit components or changes
- [X] Feature is minimally responsive
- [X] Manual testing of UI feature
- [X] Add copyrights to new files
- [X] Connected ticket to PR

### DIFF
--- a/components/pages/file-entity/FILE_ENTITY_QUERY.gql
+++ b/components/pages/file-entity/FILE_ENTITY_QUERY.gql
@@ -31,15 +31,12 @@ query FILE_ENTITY_QUERY($filters: JSON) {
           file_access
           study_id
           analysis_tools
+          embargo_stage
 
           file {
             size
             md5sum
             name
-          }
-
-          meta {
-            embargo_stage
           }
 
           repositories {
--- a/components/pages/file-entity/FILE_ENTITY_QUERY.gql
+++ b/components/pages/file-entity/FILE_ENTITY_QUERY.gql
@@ -38,6 +38,10 @@ query FILE_ENTITY_QUERY($filters: JSON) {
             name
           }
 
+          meta {
+            embargo_stage
+          }
+
           repositories {
             hits {
               edges {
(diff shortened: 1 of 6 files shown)


demariadaniel:
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/icgc-argo/platform-ui/2148

components/pages/file-entity/FILE_ENTITY_QUERY.gql
```diff
@@ -38,6 +38,10 @@ query FILE_ENTITY_QUERY($filters: JSON) {
             name
           }
 
+          meta {
+            embargo_stage
+          }
+
```
joneubank, line 44: I suppose its not clear, there is the `embargo_stage` property at the root level as well, and they should have the same value - let's use that instead, ideally we can simplify the gql schema that arranger generates to hide the duplicated version in meta. At very least lets choose one or the other to be consistent and I'd suggest the root level one.

The reason they've been duplicated is to simplify some complicated filtering rules for the api, I can break down the details of that for you if you're interested...

joneubank (approved):

components/pages/file-entity/FILE_ENTITY_QUERY.gql
demariadaniel: Oh definitely overlooked that, thanks, updated!
I think this simplifies the whole thing too
Let me know if `embargoStage` belongs in `EntityData > data` but I figure since it's in the root query it could be in the root object
And yes, breaking down the filtering rules would be interesting to hear!

joneubank (approved):

Issue #2156, opened 2022-01-12
File Repo: implement tabs for facet panel


kcullion:
**THIS SHOULD BE BEHIND A FEATURE FLAG**

## Goal
When the clinical data is indexed in the file repository, we want to display the facets in two tabbed panels, similar to: https://dcc.icgc.org/repositories
In this ticket, we will set up the tabs and arrange the facets that currently do exist in the file repository.

![image](https://user-images.githubusercontent.com/18424138/149182001-42ee7ed6-627c-4961-88ca-e10e1311f067.png)

## Expected outcomes
- [ ] Tabs are styled as per mockup, and working to switch facet list

Active and default tabs shown: Zeplin: https://zpl.io/VkL7x8G
Hover state just needs to be on the default state tab (not the active) seen here on the File Filters tab (just change background colour to #F2F2F8)
![image](https://user-images.githubusercontent.com/18424138/149184239-19b94541-5584-41a9-b904-44a5fa8e9bac.png)

Note these tabs look a bit different than the ones that exist in storybook right now - those are not noticeable enough: https://argo-ui-storybook.netlify.app/?path=/story/uikit-tabs--basic
You could use these as a starting point and make a "rounded tabs" component (they could be tabs we want to reuse in the future)

### Clinical Filters tab
- [ ] For now, the clinical tabs can show the facets that exist in the current file repository (we will hook the rest up in this ticket:https://github.com/icgc-argo/platform-ui/issues/2158)
`Program ID`
`Specimen Type`
`Specimen Tissue Source`
Note: they should be open facets by default - they appear closed in the mockup just to make it easier for design purposes.
![image](https://user-images.githubusercontent.com/18424138/149193219-de230044-ce1b-4df1-8220-63e1667fe508.png)

## File Filters tab
- [ ] All of these facets currently exist in the file repository - you can display them in the file filters tab as shown in in the mockup. Again, all facets should be open by default
Zeplin: https://zpl.io/bAPD0Kp

## Access Tier facet
- [ ] when a program member or dcc admin are logged in, sometimes an access tier facet is displayed. This will display in both clinical and file filter tabs, below the search facet, as shown in: https://zpl.io/bAPD0Kp
![image](https://user-images.githubusercontent.com/18424138/149188084-19626daf-f1fc-401e-a90c-0cf85e0ec8d1.png)

demariadaniel:
@kcullion In addition to Access Tier does release state display on both tabs?

[...]

PR #2167, opened 2022-01-19
✨ #2156 Facet Panel Tabs

demariadaniel:
**Description of changes**

Implements Tabs for toggling between Clinical Filters and File Filters on the Facet Panel
Feature flagged

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
  - [x] Modified Existing components and updates stories
- [x] Connected ticket to PR


components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -448,25 +469,40 @@ const FacetPanel = () => {
...
   });
+
+  const [currentTab, setTabs] = useState('clinical');
+  const tabStyles = (tabValue) => css`
+    align-items: center;
+    justify-content: center;
+    padding: 5px 20px;
+    margin: 9px 5px 0px;
+    border: 1px solid ${theme.colors.grey_2};
+    border-top: ${currentTab === tabValue ? `4px solid ${theme.colors.secondary}` : null};
```
samrichca, line 480: I would use `0` instead of `null` here. I don't think `null` will break anything per se, but it's not a standard CSS value.

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -448,25 +469,40 @@ const FacetPanel = () => {
...
-            `}
-            color={theme.colors.primary}
-          >
-            Filters
-          </Typography>
-        </FacetRow>
+        {FEATURE_FACET_TABS_ENABLED && (
+          <FacetRow>
+            <Tabs value={currentTab} onChange={(e, value) => setTabs(value)}>
+              <Tab value="clinical" label="Clinical Filters" css={tabStyles('clinical')} />
```
samrichca, line 501: the CSS prop here. nice!

demariadaniel: thanks i wasn't sure how this would be received as the rest of the app is largely using css inline ;)

components/pages/file-repository/FacetPanel/index.tsx
demariadaniel: good point, maybe i can use 'initial' or 'inherit', i was trying to 'short circuit' here so there are no changes unless it's the active tab

demariadaniel: OK turns out the whole thing could be greatly simplified by just leveraging the `active` class, give it a 2nd look when you can. thanks!

samrichca: even better. ok, ship it!

samrichca (approved):

mistryrn (changes requested):
Looks good! We should probably also add this feature flag to the `.env.schema` as well. Once I added the flag to my .env it worked like a charm!

I'm not sure the exact process to add this flag to dev/qa/prod, I think it has something to do with helm charts. Someone like Henro would likely be able to point you in the right direction there.

demariadaniel:
> Looks good! We should probably also add this feature flag to the `.env.schema` as well. Once I added the flag to my .env it worked like a charm!
> 
> I'm not sure the exact process to add this flag to dev/qa/prod, I think it has something to do with helm charts. Someone like Henro would likely be able to point you in the right direction there.

Added to `env.schema`
PR for infra dev here: https://github.com/OICR/argo-infra/pull/694

mistryrn (approved):
🚀

Issue #1794, opened 2021-01-15
🐛 User with no access should have a disabled token on user profile


rosibaj:
## Steps To Reproduce
1. Login to platform with a user with no access (No DACO access and does not belong to a program)
the user JWT looked like this: 
```
{
  "iat": 1610740810,
  "exp": 1610742610,
  "sub": "23f1d054-14d5-4443-9df0-8460193dffbc",
  "iss": "ego",
  "jti": "04c7ed72-df15-43b8-92c7-e1072b8dfa01",
  "context": {
    "scope": [
      "*.DENY"
    ],
    "user": {
      "name": "RBajari@oicr.on.ca",
      "email": "RBajari@oicr.on.ca",
      "status": "APPROVED",
      "firstName": "Rosi",
      "lastName": "Bajari",
      "createdAt": 1591120423904,
      "lastLogin": 1610740810624,
      "preferredLanguage": null,
      "type": "USER",
      "groups": []
    }
  },
  "aud": []
}
```
2. Generate a token.  It spins forever on ui: 
3. 
![image](https://user-images.githubusercontent.com/15824252/104773173-a2758180-5742-11eb-9c9b-306486d42b8c.png) 
4. these errors are observed:
```
{errors: [{message: "The 'Policy' entity with name '*' was not found", locations: [],…}], data: null,…}
data: null
errors: [{message: "The 'Policy' entity with name '*' was not found", locations: [],…}]
0: {message: "The 'Policy' entity with name '*' was not found", locations: [],…}
extensions: {timestamp: "2021-01-15T20:01:05.619+0000", status: 404, error: "Not Found",…}
locations: []
message: "The 'Policy' entity with name '*' was not found"
path: ["generateAccessKey"]
extensions: {tracing: {version: 1, startTime: "2021-01-15T20:01:05.382Z", endTime: "2021-01-15T20:01:05.629Z",…}}
tracing: {version: 1, startTime: "2021-01-15T20:01:05.382Z", endTime: "2021-01-15T20:01:05.629Z",…}
duration: 246992728
endTime: "2021-01-15T20:01:05.629Z"
execution: {,…}
startTime: "2021-01-15T20:01:05.382Z"
version: 1

```

## Expected behaviour

- [ ] Disable the "Generate" button
- [ ] Remove the "Expired" tag that shows up in the token field (this is what I saw as a basic user for the platform in production)
- [ ] For this use case only: modify the text of the orange warning banner to be the following:

[...]

PR #2193, opened 2022-02-03
🐛 #1794 Update Profile Token Generation Logic

demariadaniel:
**Description of changes**

Checks JWT for Scopes to determine whether user can generate a token

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [x] No changes to UI
- UI Kit (select one):
- [x] Connected ticket to PR

### DIFF
--- a/components/pages/user/index.tsx
+++ b/components/pages/user/index.tsx
@@ -46,7 +46,7 @@ export function UserPage({ scope }) {
   const isDacoApproved = get(data, ['self', 'isDacoApproved']);
   const apiToken = get(data, ['self', 'apiKey']);
   const programs = get(data, 'programs');
-  const hasTokenAccess = scope && scope.some((element) => element.match(/.READ|.WRITE/gm));
+  const hasTokenAccess = scope && scope.some((element) => element.match(/\.READ|\.WRITE/));
 
   return (
     <DefaultLayout>
--- a/components/pages/user/index.tsx
+++ b/components/pages/user/index.tsx
@@ -46,7 +46,7 @@ export function UserPage({ scope }) {
   const isDacoApproved = get(data, ['self', 'isDacoApproved']);
   const apiToken = get(data, ['self', 'apiKey']);
   const programs = get(data, 'programs');
-  const hasTokenAccess = scope && scope.some((element) => element.match(/\.READ|\.WRITE/));
+  const hasTokenAccess = scope && scope.some((element) => element.match(/(\.READ|\.WRITE)$/));
 
   return (
     <DefaultLayout>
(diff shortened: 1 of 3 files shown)


components/pages/user/index.tsx
```diff
@@ -41,12 +41,12 @@ const Column = (props) => (
...
 );
 
-export function UserPage() {
+export function UserPage({ scope }) {
   const { data, loading } = useQuery<ProfileQueryData>(PROFILE, { variables: {} });
   const isDacoApproved = get(data, ['self', 'isDacoApproved']);
   const apiToken = get(data, ['self', 'apiKey']);
   const programs = get(data, 'programs');
-  const hasProgramAccess = programs && programs.length > 0;
+  const hasTokenAccess = scope && scope.length > 0;
```
joneubank, line 49: unfortunately, Ego's hack for empty scopes is a JWT with scope = {*.DENY} so this would return true...

Can we check for something equivalent to "has at least one entry with READ or WRITE permission"?

There are methods in the `egoTokenUtils` that can return the individual scopes of the JWT (or you can parse this yourself).

joneubank (changes requested):

components/pages/user/index.tsx
demariadaniel: OK I've updated this to check for Read / Write scopes
& I've also removed checks for Daco approval for token generation
Take a second look when you get a chance, thanks!

components/pages/user/index.tsx
```diff
@@ -41,12 +41,12 @@ const Column = (props) => (
...
 );
 
-export function UserPage() {
+export function UserPage({ scope }) {
   const { data, loading } = useQuery<ProfileQueryData>(PROFILE, { variables: {} });
   const isDacoApproved = get(data, ['self', 'isDacoApproved']);
   const apiToken = get(data, ['self', 'apiKey']);
   const programs = get(data, 'programs');
-  const hasProgramAccess = programs && programs.length > 0;
+  const hasTokenAccess = scope && scope.some((element) => element.match(/.READ|.WRITE/gm));
```
joneubank, line 49: Nice use of regex. I think we shouldn't have global or multiline flags, instead confirm that one of these ends the line (though this is nitpicking cause in theory they always will).

Something like `/(\.READ|\.WRITE)$/`

joneubank: (also `.` checks any character, so we do at least need that escaped like `\.` to not match a policy with READ or WRITE in it)

demariadaniel: Yep all good points, updated!

joneubank (approved):
Beauty. Thank you sir.

Issue #2159, opened 2022-01-12
File Repo: new Search by Donor ID facet that returns donors


kcullion:
**THIS SHOULD BE BEHIND THE FEATURE FLAG** 
NOTE: this may need an API ticket, please investigate and if so, let us know
See ICGC facet panel for search box examples: https://dcc.icgc.org/repositories

## Goal
To split out the Search by Donor ID into a new component that is displayed in the clinical filters tab. This will allow the user to search by donor id/submitter donor id, return a list of matching donors, then to choose a donor and access all files for them.

## Expected Outcomes
- [x] Placeholder for this search box `e.g. DO35083, PCSI_0103...` 
- [x] with an info bubble that states: `Enter a Donor ID or Submitter Donor ID.`
NOTE: Ignore the "Upload a list of IDS" button for now - this is a future feature to be speced out (so don't display the button yet).
![image](https://user-images.githubusercontent.com/18424138/150192127-6e16651b-7897-409a-8b64-99260eef78dd.png)

- [x] When the user enters a donor_id or a submitter_donor_id, return a list of up to 5 donors that match the input.
The result should have the format:
```
donor_id (submitter_donor_id)
primary_site
```
Shown in Zeplin: https://zpl.io/a8dopjJ
![image](https://user-images.githubusercontent.com/18424138/150192282-48a20aba-de00-471b-bb9f-f974f12a8021.png)

## Selecting a donor result
As the mockup below shows, when the user chooses a donor result from this list:

- [x] 1. The donor_id is added above the search box with the ability to remove it. 
- [x] 2. The donor_id is added to the sqon as shown in mockup
- [x] 3. The file list and charts are filtered **on that donor_id - showing all of the files that exist for that donor**
- [x] 4. The user can still add multiple ids in this search box, which also get added to the sqon and above the search box (like you fixed for the file search)
 
![image](https://user-images.githubusercontent.com/18424138/150206757-ce654d55-7c32-4604-8dba-cac6e03cd4ba.png)

 
 ## Current Search box should now just search file ID and object id
 
- [x] Remove the ability to search by donor id/submitter donor id in the current searchbox, it will now just search by file id and object id
- [x] change the title to "Search by File ID" 

- [x] Placeholder for this search box `e.g. FL13796, 009f4750-e167...` 
- [x] with an info bubble that states: `Enter a File ID or Object ID.`
[...]

PR #2199, opened 2022-02-10
✨ #2159 Search by Donor ID

demariadaniel:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
- [x] Connected ticket to PR

### DIFF
--- a/components/pages/file-entity/AssociatedDonors/index.tsx
+++ b/components/pages/file-entity/AssociatedDonors/index.tsx
@@ -19,11 +19,10 @@
 
 import { FileCard, TableDiv } from '../common';
 import { css } from '@emotion/core';
-import VerticalTabs from 'uikit/VerticalTabs';
+// import VerticalTabs from 'uikit/VerticalTabs';
 import { Col, Row } from 'react-grid-system';
 import { DonorRecord } from '../types';
 import React from 'react';
-import Link from 'uikit/Link';
 import Typography from 'uikit/Typography';
 import styled from '@emotion/styled';
 import { useTheme } from 'uikit/ThemeProvider';
--- a/components/pages/file-repository/FacetPanel/SearchResultsMenu.tsx
+++ b/components/pages/file-repository/FacetPanel/SearchResultsMenu.tsx
@@ -20,7 +20,7 @@
 import { css, styled } from 'uikit';
 import Typography from 'uikit/Typography';
 import Icon from 'uikit/Icon';
-import { IdSearchQueryData } from './types';
+import { FacetDetails, FileFacetPath, IdSearchQueryData } from './types';
 import { useTheme } from 'uikit/ThemeProvider';
 import theme from 'uikit/theme/defaultTheme';
 
@@ -55,10 +55,12 @@ const ListItem = styled(Typography)`
 `;
 
 const SearchResultsMenu = ({
+  currentSearch,
   isLoading,
   searchData,
   onSelect,
 }: {
+  currentSearch: FacetDetails;
   isLoading: boolean;
   searchData: IdSearchQueryData;
   onSelect: Function;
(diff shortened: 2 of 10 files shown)


components/pages/file-entity/AssociatedDonors/index.tsx
```diff
@@ -19,11 +19,10 @@
 
 import { FileCard, TableDiv } from '../common';
 import { css } from '@emotion/core';
-import VerticalTabs from 'uikit/VerticalTabs';
+// import VerticalTabs from 'uikit/VerticalTabs';
```
joneubank, line 22: why comment?

components/pages/file-repository/FacetPanel/SearchResultsMenu.tsx
```diff
@@ -104,8 +112,8 @@ const SearchResultsMenu = ({
                     border-bottom: 0px;
                   }
                 `}
-                onClick={() => onSelect(node.file_id)}
-                key={node.file_id}
+                onClick={() => onSelect(nodeData)}
+                key={`${nodeData}-${i}`}
```
joneubank, line 116: 👍

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -336,10 +407,10 @@ const FacetPanel = () => {
...
   };
 
-  const { data: idSearchData, loading: idSearchLoading } = useIdSearchQuery(
-    debouncedSearchTerm,
-    excludedIds,
-  );
+  const { data: idSearchData, loading: idSearchLoading } =
+    currentTab === 'clinical'
+      ? useDonorIdSearchQuery(debouncedSearchTerm, excludedIds)
+      : useFileIdSearchQuery(debouncedSearchTerm, excludedIds);
```
joneubank, line 413: specific variables changing based on `currentTab` instead of a whole include change seems prone to maintenance issues long term... 

are there multiple places in the codebase where we switch functionality by tab? should be wrapped all in a single place.

components/pages/file-entity/AssociatedDonors/index.tsx
demariadaniel: So I meant to bring this up in standup; we have several files with commented out code, and then multiple unused imports at the top of the file. I'd prefer to clean these up, decide if we can revisit the features that were commented out or remove them.

So this line is commented out b/c it's unused, the code it relates to is also commented out

joneubank: general philosophy is to remove unused code. Unless theres a comment along side the commented out block explaining why its been left (as reference or for future use or something) then I have no concern about it being removed.

unused imports should be removed as well but I think we've been less careful about that.

components/pages/file-repository/FacetPanel/index.tsx
demariadaniel: OK I've made pretty significant refactors to remove any conditionals. The code is always checking against the properties of our current selected search function. Please review changes to FacetPanel & SearchResultsMenu (plus a few type updates)

I also needed a few updates so that the Donor search does not display duplicate results, which required some pretty intense nested filtering logic.

Some of the variable names could likely be improved, i.e. `primary`, `secondary` and `tertiary` represent search-menu content, so something like `searchResultText`, `resultSecondaryText` and `resultSubtitleText` might be better. 

Let me know if you have suggestions, I'll probably have better names on Monday morning ;)

components/pages/file-entity/AssociatedDonors/index.tsx
demariadaniel: this one has a TODO with a reference for the future, I can revert the commented out line for now

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -266,6 +316,54 @@ const useIdSearchQuery = (
...
+          !filteredIds.includes(hit.node.submitter_donor_id)
+        ) {
+          filteredIds.push(hit.node.donor_id, hit.node.submitter_donor_id);
+
+          searchResults.push({
+            resultId: hit.node.donor_id,
+            secondaryText: hit.node.submitter_donor_id,
+            subText: '',
+          });
+        } else return false;
```
mistryrn, line 360: What does returning `false` in a `forEach` loop do? Is it equivalent to `continue` in a regular for loop?

mistryrn (commented):
Looks good so far! Just the one comment about that `return false` in a `forEach`

components/pages/file-repository/FacetPanel/index.tsx
demariadaniel: Found this will exit the loop (equivalent to a `break`) so I've removed this line

PR #2233, opened 2022-03-09
Sort fileId column using fileNumber

joneubank:
**Description of changes**
Uses numeric value of the file id for sorting the file ID column in the file repository.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [x] No design provided, screenshot of changes included in PR
  - [ ] No changes to UI
- UI Kit (select one):
- [x] Connected ticket to PR

### DIFF
--- a/components/pages/file-repository/FileTable/index.tsx
+++ b/components/pages/file-repository/FileTable/index.tsx
@@ -60,7 +60,7 @@ const DEFAULT_PAGE_SIZE = 20;
 const DEFAULT_PAGE_OFFSET = 0;
 const DEFAULT_SORT = [
   {
-    field: FileCentricDocumentField.file_id,
+    field: FileCentricDocumentField.file_number,
     order: 'asc' as FileRepositoryRecordSortOrder,
   },
 ];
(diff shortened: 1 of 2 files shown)


components/pages/file-repository/FileTable/index.tsx
```diff
@@ -60,7 +60,7 @@ const DEFAULT_PAGE_SIZE = 20;
 const DEFAULT_PAGE_OFFSET = 0;
 const DEFAULT_SORT = [
   {
-    field: FileCentricDocumentField.file_id,
+    field: FileCentricDocumentField.file_number,
     order: 'asc' as FileRepositoryRecordSortOrder,
```
demariadaniel, line 64: I think we actually want this to be 'desc'
When I test on localhost, 'asc' starts at FL705 and goes up
'desc' starts at FL3894 and goes down
AFAIK I think we want newest files first i.e. highest file number

joneubank: good catch, will adjust.

demariadaniel (approved):

PR #2237, opened 2022-03-11
🐛 Fix crash when `metrics` is `null`

mistryrn:
**Description of changes**

* Use lodash `isObject` to check type instead of relying on `typeof`, preventing a crash where the UI would try to do an `Object.values()` call on a null metrics object

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [x] No changes to UI
- UI Kit (select one):
- [x] Connected ticket to PR

### DIFF
--- a/components/pages/file-entity/useEntityData.tsx
+++ b/components/pages/file-entity/useEntityData.tsx
@@ -18,7 +18,12 @@
  */
 
 import { useQuery } from '@apollo/react-hooks';
-import { get, mapKeys, camelCase } from 'lodash';
+import {
+  camelCase,
+  get,
+  isObject,
+  mapKeys,
+} from 'lodash';
 import {
   FileSummaryInfo,
   FileAccessState,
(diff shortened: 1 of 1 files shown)


components/pages/file-entity/useEntityData.tsx
```diff
@@ -43,7 +48,7 @@ const noData = { programShortName: null, access: null, size: null, data: null, e
 
 const isValidMetricsObject = (metrics: any) => {
   return (
-    typeof metrics === 'object'
```
mistryrn, line 46: For some reason this `typeof` check wasn't short-circuiting when the value passed in was `null`, causing a crash when checking the following two conditions. `isObject` seems to be stricter, not getting the crash with this now.

joneubank (approved):

PR #2272, opened 2022-03-30
🧪 2126 Primary Diagnosis tab

demariadaniel:
**Description of changes**

https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/icgc-argo/platform-ui/2126

* Formats Primary Diagnosis entity data for display on Clinical Timeline
* Sets interval to 0 & displays Age flag correctly for primary diagnosis & deceased scenarios
* Displays correct labels for PRIMARY DIAGNOSIS, submitter_primary_diagnosis_id and cancer_type_code  
* Displays data as two tables
* Aliases clinical categories as Clinical TNM Category
* Handles Pipe delimiter in data
* Adds multiple display utility functions & refactors data handling

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
- [x] Connected ticket to PR


components/pages/donor-entity/ClinicalTimeline/Timeline.tsx
```diff
@@ -171,6 +171,15 @@ const Timeline = ({
   const theme = useTheme();
   const timelineStyles = React.useMemo(() => getTimelineStyles(theme), [theme]);
 
+  const donorAgeDisplay = (type, data) => {
```
ciaranschutte, line 174: (type: EntityType, data: any) => ...
little sprinkle of TS.. not too sure what data is..

components/pages/donor-entity/ClinicalTimeline/util.tsx
```diff
@@ -114,47 +146,109 @@ const donorCentricDisplayNames = {
...
 
-  const ageAtDeath = ageAtDiagnosis + survivalTime;
+  const ageAtDeath: number = ageAtDiagnosis + survivalTime;
   return { ageAtDiagnosis, survivalTime, ageAtDeath };
 };
 
-type AliasedDisplayData = {
-  [K in keyof typeof donorCentricDisplayNames]?: any;
+const removePipeDelimiter = (data: AliasedDisplayData) => {
+  let displayData = { ...data };
```
ciaranschutte, line 222: i think this can be a const too because you're just reassigning keys
const vs let is memory level. so I can have a have a `const obj = {a: 1, b: 2}` and then change `obj.a = 2` because its still that object in memory. but I couldn't do `obj = {a: 2, b:2}` because im assigning a new value of a new object

components/pages/donor-entity/index.tsx
```diff
@@ -32,11 +32,12 @@ import DonorCardsLayout from './DonorCardsLayout';
...
-  const donorLoading = false;
-
   const { egoJwt } = useAuthContext();
   const { data: userProfile, loading: profileLoading } = useQuery(USER_PROFILE);
+  const {
+    loading: donorLoading,
+    summary: { program_id, donor_id },
+  } = donor;
 
   const loading = profileLoading || donorLoading;
```
ciaranschutte, line 43: an error boundary somewhere (probably here) would be beneficial.
we are accounting for loading but not for error of no data/bad data

components/pages/donor-entity/ClinicalTimeline/util.tsx
```diff
@@ -114,47 +146,109 @@ const donorCentricDisplayNames = {
...
   workflow_names: 'Workflow Names',
 };
 
+type AliasedDisplayData = {
+  [K in keyof typeof donorCentricDisplayNames]?: any;
+};
+
 // format for display
 export const formatTableDisplayNames = (data: any[]) =>
   data.length > 0 &&
```
ciaranschutte, line 171: data && data.length > 0 
if this is called with no argument data.length will cause type error

components/pages/donor-entity/ClinicalTimeline/util.tsx
demariadaniel: Updated!

components/pages/donor-entity/ClinicalTimeline/Timeline.tsx
demariadaniel: thanks for the sprinkles! ;)
yeah this is a tricky one b/c each 'entity' has a different shape for `data`
could be a treatment node, primary diagnosis, specimen, etc., so `data` is basically a union type of many different types
But until I work through all the different Timeline Tab entities and fully update the dummy data, using a union type like that would throw a TS error
so I'll leave `data` blank for now and update that in a future timeline ticket (which could be next sprint anyways)

components/pages/donor-entity/index.tsx
demariadaniel: this is something I've struggled with, a lot of the JS query handling seems to be written as if we expect to never receive/handle errors. `useEntityData` does include handling for `!entity` and deconstructs `error` from the query response so I'll start there and see what I can update.

ciaranschutte: yes we're all very optimistic developers ;)

demariadaniel: Hey I looked into this again and `useEntityData` is pretty similar in `donor-entity` and `file-entity`, and we're basically handling this upstream

If there's no `entity` query data returned, we return the `noData` object; if there's an error we print it to the console
And in `[donorId].tsx` we only render the DonorEntityPage if we have a valid entity, but the logic there is still using testing data until the API is up

Do you think that's goo enough? anything we should update?

ciaranschutte: I think that's definitely good enough in regards to this PR. consistent. thanks for checking that.

components/pages/donor-entity/ClinicalTimeline/util.tsx
demariadaniel: Thanks, updated w/ `data &&`

ciaranschutte (approved):
great stuff!

PR #2326, opened 2022-06-07
🎨 💬 2318: update announcements

samrichca:
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/icgc-argo/platform-ui/2318

# Description of changes

- updated Announcements
- refactor component to make HTML more semantic and DRY
  - add `<h3>`, use `<p>` inside `<div>`
  - resolve console error about `<div>` inside a `<p>`

## Type of Change

- [ ] Bug
- [x] Styling
- [ ] New Feature

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- UI Kit (select one):
  - [ ] Modified Existing components
- [x] Connected ticket to PR

### DIFF
--- a/components/pages/Homepage/common.tsx
+++ b/components/pages/Homepage/common.tsx
@@ -17,6 +17,7 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+import { ReactNode } from 'react';
 import { css, styled } from 'uikit';
 import { useTheme } from 'uikit/ThemeProvider';
 import Typography from 'uikit/Typography';
@@ -34,7 +35,7 @@ import { reactGridBreakpoints } from 'global/utils/common';
 
 export type NewsItem = {
   title: string;
-  text: string | JSX.Element;
+  text: ReactNode;
 };
 
 type DataReleaseStatistic = {
@@ -302,7 +303,7 @@ export const NewsContainer: React.ComponentType<{ newsItems: NewsItem[] }> = ({ 
               >
                 Announcements:
               </Typography>
-              {newsItems.map((newsItem) => (
+              {newsItems.map((newsItem: NewsItem) => (
                 <Typography
                   key={`newsItem-${encodeURI(newsItem.title)}`}
                   variant="paragraph"
--- a/components/pages/Homepage/common.tsx
+++ b/components/pages/Homepage/common.tsx
@@ -311,7 +311,7 @@ export const NewsContainer: React.ComponentType<{ newsItems: NewsItem[] }> = ({ 
                     margin: 1em 0 0 0;
                   `}
                 >
-                  {newsItem.title && <strong>{newsItem.title}:</strong>} {newsItem.text}
+                  {newsItem.title && <b>{newsItem.title}:</b>} {newsItem.text}
                 </Typography>
               ))}
             </div>
--- a/components/pages/Homepage/common.tsx
+++ b/components/pages/Homepage/common.tsx
@@ -303,9 +303,9 @@ export const NewsContainer: React.ComponentType<{ newsItems: NewsItem[] }> = ({ 
               >
                 Announcements:
               </Typography>
-              {newsItems.map((newsItem: NewsItem) => (
+              {newsItems.map((newsItem: NewsItem, index: number) => (
                 <Typography
-                  key={`newsItem-${encodeURI(newsItem.title)}`}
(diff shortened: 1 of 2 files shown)


components/pages/Homepage/common.tsx
```diff
@@ -280,18 +285,32 @@ export const NewsContainer: React.ComponentType<{ newsItems: JSX.Element[] }> =
...
+                  key={`newsItem-${encodeURI(newsItem.title)}`}
                   variant="paragraph"
                   css={css`
-                    padding-bottom: 15px;
-                    border-bottom: 1px solid ${theme.colors.grey_2};
+                    margin: 1em 0 0 0;
                   `}
                 >
-                  {item}
+                  {newsItem.title && <strong>{newsItem.title}:</strong>} {newsItem.text}
```
ciaranschutte, line 313: semantically i don't think it makes sense to have every news item title be `<strong/>`. it's saying every single title has a strong importance in the page.

if it's the default bold styling you're after please use a bold tag or I think `<Typography>` bold prop may be sufficient. Failing that something working with what we have already in typography.

components/pages/Homepage/common.tsx
```diff
@@ -280,18 +285,32 @@ export const NewsContainer: React.ComponentType<{ newsItems: JSX.Element[] }> =
...
+                css={css`
+                  margin: 0;
+                `}
+              >
+                Announcements:
+              </Typography>
+              {newsItems.map((newsItem) => (
                 <Typography
-                  key={`newsText-${index}`}
+                  key={`newsItem-${encodeURI(newsItem.title)}`}
```
ciaranschutte, line 307: is there a specific use case for encoding the title rather than just using the `index` from `map`?
key just needs to be diff for react algo. this introduces the possibility of having the same title.
more importantly a mental step of why are we encoding the title...

components/pages/Homepage/common.tsx
```diff
@@ -32,6 +32,11 @@ import scientistImage from 'static/icgc-data-scientist-wide.jpg';
 import DnaLoader from 'uikit/DnaLoader';
 import { reactGridBreakpoints } from 'global/utils/common';
 
+export type NewsItem = {
+  title: string;
+  text: string | JSX.Element;
```
ciaranschutte, line 37: looks like `text` fields are React fragments. `ReactNode` is more applicable than `JSX.Element` which I believe is more of a generic type for other transformers/applications. our type is just here's some react display stuff... please show...

components/pages/Homepage/common.tsx
```diff
@@ -280,18 +285,32 @@ export const NewsContainer: React.ComponentType<{ newsItems: JSX.Element[] }> =
               padding: 20px;
             `}
           >
-            <div>
-              {/** @Note DO NOT REMOVE div. Magic fix needed here to enforce style  */}
```
ciaranschutte, line 284: noooo not the magic div /s

ciaranschutte (changes requested):
couple of q's / suggestions. looks good.

components/pages/Homepage/common.tsx
ciaranschutte: which means you can change it to just `ReactNode` ... string is covered by that

components/pages/Homepage/common.tsx
samrichca: You shouldn't use the index from map for keys AFAIK, but I figured the titles would be unique, but they might have spaces. I'll think about it and come back to this

components/pages/Homepage/common.tsx
samrichca: string is in fact covered by ReactNode, good call

components/pages/Homepage/common.tsx
ciaranschutte: not a big deal. seems like react defaults to using indexes.
you're right - it's not a good idea because what if you remove one and now the indexes are different. 

this is definitely not a big deal here as it's a static news section (nothing changing at runtime)
a little fyi in #frontend may be worthwhile. we use index as key **everywhere** :p

samrichca: Since it's static content I'm changing it to index 👍

samrichca: Anything we get from an API should have some kind of ID associated with it and that's what we should use for the key. We do use index everywhere so I was overcorrecting

components/pages/Homepage/common.tsx
samrichca: Noted & updated

Issue #2309, opened 2022-05-20
Upgrade Next 10 to 11, Webpack 4 to 5


samrichca:
Upgrade Next 10 to 11, which requires Webpack 5

b-f-chan:
Question @samrichca - There's also a ticket to upgrade Next from 11 to 12, does this need to be done sequentially or can you skip from 10 to 12?  @samrichca to advise

b-f-chan:
Team decision --> Hold back Epic #2208 from Loki to ensure we have enough time to test thoroughly

b-f-chan:
Some remaining issues need to be resolved, moving back to In Progress --> Need to change how we import images and various error pages

samrichca:
https://github.com/icgc-argo/platform-ui/pull/2360

PR #2360, opened 2022-07-20
#2309 - update Next 10 to 11

samrichca:
# Description of changes

- update Next 10 to 11
- add `React.Fragment`
- move `/static` to `/public/static` due to deprecation of `/static`
  - related to this: change images from imports to direct references
- update webpack config due to API changes in webpack 5

## Type of Change

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [x] No changes to UI
- [x] Connected ticket to PR

https://github.com/icgc-argo/platform-ui/issues/2309


next-env.d.ts
```diff
@@ -1,39 +1,6 @@
-/*
```
samrichca, line 1: this file gets regenerated. if I add the copyright it gets removed.

ciaranschutte: I'm sure we added this code for a reason. Does everything still type-check correctly?
Offhand I don't know of any changes we've made to typescript/next packages that would make these additions redundant.

The nextjs docs points out we can keep this functionality by adding an additional .d.ts file
```
Instead of editing next-env.d.ts, you can include additional types by adding a new file e.g. additional.d.ts and then referencing it in the [include](https://www.typescriptlang.org/tsconfig#include) array in your tsconfig.json.
```

ciaranschutte: also if we're going by their recommended best practices this file should not be committed and should be ignored from version control
```
A file named next-env.d.ts will be created in the root of your project. This file ensures Next.js types are picked up by the TypeScript compiler. You cannot remove it or edit it as it can change at any time. As such this file should not be committed and should be ignored by version control.
```

next.config.js
```diff
@@ -37,10 +37,9 @@ module.exports = withPlugins([withTM, withImages], {
         }
       : defaultPathMap,
   webpack: (config, options) => {
-    // Fixes npm packages that depend on `fs` module
-    config.node = {
-      fs: 'empty',
-    };
+    if (!options.isServer) {
```
ciaranschutte, line 40: excellent. looks cleaner too.

next-env.d.ts
samrichca: good catch! I'll update this shortly

ciaranschutte:
Is there a reason in changing the fragment shorthand, `<>`, to the full '<React.Fragment>`?

samrichca:
@ciaranschutte fragments weren't working but it looks like all I have to do is import react, I don't have to add React.Fragment everywhere.

ciaranschutte:
> @ciaranschutte fragments weren't working but it looks like all I have to do is import react, I don't have to add React.Fragment everywhere.

This branch has react 16. I think you upgraded to 17 as part of this upgrade process?
Once you're on react 17 fragments should just work as is because it automatically does the jsx transforms without needing a React import

components/pages/403.tsx
```diff
@@ -25,8 +25,6 @@ import HyperLink from '@icgc-argo/uikit/Link';
 import Link from 'next/link';
 import { getConfig } from 'global/config';
 import { Row, Col } from 'react-grid-system';
-import image from 'static/dna-locked.svg';
```
ciaranschutte, line 28: Are the changes to the static folder required?

I just saw the part about a `public` folder in the migration guide from next 10 - 11
I think our "static" folder is poorly named potentially as it seems like that's our webserver public folder
but It doesn't look like we're using an image "statically" anywhere.
ie. all usage is an import

a downside of moving these into webserver static assets instead of build time imports is we loose the ability to process them. eg. svg inline 
also I think Nextjs image capabilities suffer a little as it can't statically process an asset if it's not imported

samrichca: I gotcha. I'm making the following changes:
- Favicon is going in the public folder
- Most other images are being loaded with the new Next `<Image />` component, and stored in `/assets/images` (new) and imported.

I was having trouble with build-time imports because of a line in the config that allows us to have absolute imports for `/components`, `/global`, other top-level folders. That breaks the built-in behaviour for `/public`. I added a comment about this in the config.

ciaranschutte: Great! If we're using the new <Image/> component can it still do all it's cool optimization stuff if we're loading statically from a static folder vs doing a build time import?

samrichca: Interestingly I found that it won't inline SVGs if I do that, but that might be a conflict with another library, so I'm testing things out

samrichca:
"This branch has react 16. I think you upgraded to 17 as part of this upgrade process?" @ciaranschutte where do you see React 16? I see 17 in package.json

ciaranschutte:
> "This branch has react 16. I think you upgraded to 17 as part of this upgrade process?" @ciaranschutte where do you see React 16? I see 17 in package.json

double checked. yeah not sure where I was seeing this. odd we still need to import React.

ciaranschutte (approved):

Issue #2268, opened 2022-03-29
[Data Hookup] Submitted data: quick filters


kcullion:
Make the quick filters items work:

- [x] 1. Show all donors - always takes the user back to the full submitted data section
- [x] 2.Show invalid donors - this item should only show if there are invalid donors (i.e there is some submitted data that doesn't comply with the most recent dictionary). When it's available it will filter the section to only show the tables with invalid rows: like https://zpl.io/r17Y8gD
- [x] 3. Show clinically complete donors - this will filter submitted data for all donors that are clinically complete
- [x] 4. Show clinically incomplete donors - this will filter submitted data for all donors that are clinically incomplete
- [x] 5. The title can sometimes say "donor_id" (if filtered on a specific donor), or "# donors" if filtered on multiple donors like in: https://zpl.io/ll48jy5

![image](https://user-images.githubusercontent.com/18424138/160695228-ab85dfb0-f4a9-46ba-a7b8-b6b3dba32525.png)

demariadaniel:
'# of Donors' title state to be handled in #2269 https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/icgc-argo/platform-ui/2269

Current branch updates title to display 'All Donors', 'Invalid Donors', 'Clinically Complete Donors' or 'Clinically Incomplete Donors'

Buwujiu:
@demariadaniel Tested in QA on DATA-CA. 
When filtered on "incomplete donors", example: https://platform-ui.qa.argo.cancercollaboratory.org/submission/program/DATA-CA/clinical-data?tab=donor,  for donors such as DO263260, DO263256, they have all green check marks on all clinical core entities, this means they are core complete. However, they show up when quick filter is set to "incomplete donors". It's likely that the calculation of `coreCompletionPercentage` in clinical api is off. Will create another ticket to investigate. 

Other than this, the filter is working as expected.

PR #2378, opened 2022-09-02
📁 #2268 Submitted Data Quick Filters

demariadaniel:
# Description of changes

* Updates Submitted Data display text + query filter based on completion state dropdown value

Argo-Clinical PR: https://github.com/icgc-argo/argo-clinical/pull/806

## Type of Change

- [ ] Bug
- [ ] Styling
- [x] New Feature

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- [x] Connected ticket to PR

### DIFF
--- a/components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
+++ b/components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
@@ -22,6 +22,7 @@ import { useState } from 'react';
 import Container from '@icgc-argo/uikit/Container';
 import DropdownButton from '@icgc-argo/uikit/DropdownButton';
 import { useTheme } from '@icgc-argo/uikit/ThemeProvider';
+import { CompletionStates } from '../common';
 import {
   searchBackgroundStyle,
   searchTitleParentStyle,
@@ -49,7 +50,7 @@ export default function SearchBar({
   onChange,
 }: {
   noData: boolean;
-  onChange: (state: string) => void;
+  onChange: React.Dispatch<React.SetStateAction<CompletionStates>>;
 }) {
   const theme = useTheme();
 
@@ -88,19 +89,19 @@ export default function SearchBar({
             menuItems={[
               {
                 display: 'Show all donors',
-                value: 'all',
+                value: CompletionStates['all'],
               },
               {
                 display: 'Show invalid donors',
-                value: 'invalid',
+                value: CompletionStates['invalid'],
               },
               {
                 display: 'Show clinically complete donors',
-                value: 'complete',
+                value: CompletionStates['complete'],
               },
               {
                 display: 'Show clinically incomplete donors',
-                value: 'incomplete',
+                value: CompletionStates['incomplete'],
               },
             ]}
           >
--- a/components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
+++ b/components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
@@ -45,21 +45,28 @@ import Button from '@icgc-argo/uikit/Button';
 import { css } from '@icgc-argo/uikit';
 import Typography from '@icgc-argo/uikit/Typography';
 
+const titleCopy = {
+  all: 'All Donors',
+  invalid: 'Invalid Donors',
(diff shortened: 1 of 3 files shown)


components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
```diff
@@ -74,29 +111,20 @@ export default function SearchBar({ noData }: { noData: boolean }) {
...
-                value: 'all',
-              },
-            ]}
+            onItemClick={(e) => {
+              onChange(e.value);
+              const completionDisplayText =
+                menuItems.find((item) => {
+                  return item.value === e.value;
+                })?.display || '- Select an option -';
+              setDisplayText(completionDisplayText);
```
ciaranschutte, line 123: I often do these kind of "menus" with a map of menu items instead of an array. Then you can just do a key lookup. Keep `e.value` as a `CompletionStates` value, then just do `menuItems[e.value]` instead of a `find` operation on every click.

ciaranschutte (approved):
Looks good. Two small things.
1. I don't think `find` gives us anything we can't do better by using a map (simplifies + quicker)
2. some stylistic things but I'll leave that up to @demariadaniel + @samrichca to decide on. eg. `const titleCopy = ...` for a constant basic map instead of `const TitleCopy` or const `TITLE_COPY`. Certainly not saying we've been consistent so far, but it would be worth trying to be consistent moving forward.

components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
ciaranschutte: Then you can do something like `Object.entries(menuItems).forEach(item => item[1])..` for displaying the items

demariadaniel: Sounds good! I'm often doing quick 'make it work' code on my first pass and revisit optimizations when I can clear my head a bit and not have to think about functionality

Had considered the map approach too, may be able to combine `titleCopy` and `menuItems` into a single map that can be used for both purposes
Also will take consistency into account, definitely think consistency makes a codebase easier to use & maintain

demariadaniel: @ciaranschutte this is updated if you want to give a 2nd look!

components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
```diff
@@ -44,16 +45,47 @@ import Button from '@icgc-argo/uikit/Button';
 import { css } from '@icgc-argo/uikit';
 import Typography from '@icgc-argo/uikit/Typography';
 
-export default function SearchBar({ noData }: { noData: boolean }) {
+const COMPLETION_OPTIONS = {
+  all: {
```
samrichca, line 49: this looks great!

components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
```diff
@@ -74,29 +106,19 @@ export default function SearchBar({ noData }: { noData: boolean }) {
...
-              },
-              {
-                display: 'Show clinically incomplete donors',
-                value: 'all',
-              },
-            ]}
+            onItemClick={(e) => {
+              onChange(e.value);
+              const completionDisplayText =
+                (COMPLETION_OPTIONS[e.value] && `Show ${COMPLETION_OPTIONS[e.value].display}`) ||
```
samrichca, line 115: I'd use a ternary here, that's basically what you wrote but it threw me for a sec.

components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
demariadaniel: Thanks refactored in like 5 minutes haha

components/pages/submission-system/program-submitted-data/SearchBar/index.tsx
demariadaniel: Done

Issue #2421, opened 2022-10-21
Bug: Mobile nav button not working


samrichca:
On dev & QA, the mobile nav buttons aren't working.
![Screen Shot 2022-11-23 at 10 07 38 AM](https://user-images.githubusercontent.com/8744047/203581274-b9f20f66-b576-4cb8-8b9b-425a6be1768d.png)

PR #2429, opened 2022-11-01
#2421: fix mobile nav button

samrichca:
# Description of changes

- add `e.stopPropagation()` so nav buttons will work
- change divs to buttons because they're interactable

## Type of Change

- [x] Bug
- [ ] Dependency updates
- [ ] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [x] No back-end changes
- [x] Manually tested changes
- [x] Added copyrights to new files
- [x] Connected ticket to PR

### DIFF
--- a/components/NavBar.tsx
+++ b/components/NavBar.tsx
@@ -48,13 +48,31 @@ import { get } from 'lodash';
 import Image from 'next/image';
 import NextLink from 'next/link';
 import queryString from 'query-string';
-import { useState, useEffect, useMemo, createRef, RefObject } from 'react';
+import { useState, useEffect, useMemo, createRef, RefObject, PropsWithChildren } from 'react';
 
 import { useScreenClass } from 'react-grid-system';
 import urlJoin from 'url-join';
 import ModalPortal from './Modal';
 import useFiltersContext from './pages/file-repository/hooks/useFiltersContext';
 import ProgramServicesModal from './pages/Homepage/ProgramServicesModal';
+
+const TransparentButton = ({
+  children,
+  onClick,
+}: PropsWithChildren<{ onClick: (e: any) => void }>) => (
+  <button
+    css={css`
+      background: transparent;
+      border: 0 none;
+      margin: 0;
+      padding: 0;
+    `}
+    onClick={onClick}
+    type="button"
+  >
+    {children}
+  </button>
+);
 
 const NavBarLoginButton = () => {
   const { asPath: path, query } = usePageContext();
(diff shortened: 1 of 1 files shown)


components/NavBar.tsx
```diff
@@ -48,14 +48,32 @@ import { get } from 'lodash';
...
+      margin: 0;
+      padding: 0;
+    `}
+    onClick={onClick}
+    type="button"
+  >
+    {children}
+  </button>
+);
+
```
joneubank, line 76: Should this be in uikit? maybe as a Button variant?

joneubank (approved):
Code's fine but maybe should utilize uikit? Consider moving it there if that makes sense.

components/NavBar.tsx
samrichca: I was thinking about it but we've never needed it before. wouldn't hurt though

joneubank: New things are new because they weren't needed before. Will it be needed again later? ;)

components/NavBar.tsx
```diff
@@ -48,14 +48,32 @@ import { get } from 'lodash';
...
 import { useScreenClass } from 'react-grid-system';
 import urlJoin from 'url-join';
 import ModalPortal from './Modal';
 import useFiltersContext from './pages/file-repository/hooks/useFiltersContext';
 import ProgramServicesModal from './pages/Homepage/ProgramServicesModal';
 
+const TransparentButton = ({
+  children,
+  onClick,
+}: PropsWithChildren<{ onClick: (e: any) => void }>) => (
```
ciaranschutte, line 62: as a team let's be consistent with not using `any` in typescript. @joneubank @samrichca  It's really a cope out just to pass the type check settings. 3 seconds search tells me this ` React.MouseEvent<HTMLButtonElement>` is probably sufficient

joneubank: yes, thanks, nice catch.

samrichca:
@joneubank @ciaranschutte I removed the TransparentButton and used FocusWrapper from Uikit which is an HTML button with browser-default styling removed

ciaranschutte:
@samrichca code is great! is there someway we could our process to catch this earlier? I don't remember hearing about this TransparentButton ticket in sprint planning and I could have shared that the FocusWrapper exists already.

samrichca:
@ciaranschutte in this case no, the ticket was for the button not working. changing it from a div to a button was extra

ciaranschutte (approved):

PR #2604, opened 2023-03-21
Mergeback rc/1.21.1

samrichca:
Mergeback rc/1.21.1


daniel-cy-lu:
I didn't make changes to platform UI this sprint. :)

daniel-cy-lu (approved):
QA merge, version update in package.json.

demariadaniel:
@daniel-cy-lu for item #4, we actually use Semantic Versioning to determine the latest version #, based on which PR's are included in the release: https://semver.org/

For example, the latest release branch is 1.121.1, because it is only patching minor bugs, vs 1.122.0, which would be adding minor, backwards compatible features

daniel-cy-lu:
@demariadaniel I see! thanks for the link, Dan. Right, the release we are doing this time will have an update in the patch number (the right most number) and when Sam and I did it last time, it must have been a minor update and not just bug fixes so we did an update in the middle or the minor number! Got cha! I will update the documentation!

PR #2616, opened 2023-03-29
remove unmatched id field and update matched id calculation

daniel-cy-lu:
The unmatched id field has a vague definition and unknown usefulness to the users, so it's removed.

The matched id field now returns all the id that is matched. There are two cases to think about:

1. When the exact same id is enter more than once -> it will be count as 1. 
    This is same as before as we don't want to double count duplicates.
2. When the donor and submitter id from the same entry is entered -> it will be count as 2.
     BA in the ticket stated the goal is to allow the users to know the total number of matches.
3. When the donor and submitted id from the same entry is entered more than once -> due to 1. this will still be counted as 2 
     because 1. eliminates duplicates. 

## Type of Change

- [ ] Bug
- [ ] Dependency updates
- [ ] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [ ] No back-end changes
- [ ] Manually tested changes
- [ ] Connected ticket to PR

### DIFF
--- a/components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/MatchResults.tsx
+++ b/components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/MatchResults.tsx
@@ -20,48 +20,23 @@
 import React from 'react';
 import { css } from '@icgc-argo/uikit';
 
-export default function MatchResults({
-  numMatched,
-  numUnmatched,
-}: {
-  numMatched: number;
-  numUnmatched: number;
-}) {
+export default function MatchResults({ numMatched }: { numMatched: number }) {
   return (
-    <>
-      <div
+    <div
+      css={css`
+        display: flex;
+        align-items: center;
+      `}
+    >
+      Matched IDs:
+      <b
         css={css`
-          display: flex;
-          align-items: center;
+          margin-left: 4px;
+          color: #0774d3;
         `}
       >
-        Matched IDs:
-        <b
-          css={css`
-            margin-left: 4px;
-            color: #0774d3;
-          `}
-        >
-          {numMatched}
-        </b>
-      </div>
-      {/* <div> of unmatched ID */}
-      <div
-        css={css`
-          display: flex;
-          align-items: center;
-        `}
-      >
-        Unmatched IDs:
-        <b
-          css={css`
-            margin-left: 4px;
-            color: #0774d3;
-          `}
-        >
-          {numUnmatched}
-        </b>
-      </div>
-    </>
+        {numMatched}
+      </b>
+    </div>
   );
 }
--- a/components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/index.tsx
+++ b/components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/index.tsx
@@ -61,7 +61,6 @@ export default function FilterModal({
 
   //make matchID dynamic using useState
   const [numMatched, setNumMatched] = useState(0);
-  const [numUnmatched, setNumUnmatched] = useState(0);
   const [matchedIds, setMatchedIds] = useState('');
 
   // Match text area contents for Donor ID #s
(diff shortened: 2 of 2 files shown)


components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/MatchResults.tsx
```diff
@@ -20,48 +20,23 @@
...
-            margin-left: 4px;
-            color: #0774d3;
-          `}
-        >
-          {numMatched}
-        </b>
-      </div>
-      {/* <div> of unmatched ID */}
-      <div
-        css={css`
```
daniel-cy-lu, line 50: remove the `div` that holds the unmatched id field UI

components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/index.tsx
```diff
@@ -61,7 +61,6 @@ export default function FilterModal({
 
   //make matchID dynamic using useState
   const [numMatched, setNumMatched] = useState(0);
-  const [numUnmatched, setNumUnmatched] = useState(0);
   const [matchedIds, setMatchedIds] = useState('');
 
   // Match text area contents for Donor ID #s
```
daniel-cy-lu, line 66: remove the `useState` of numUnmatched

components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/index.tsx
```diff
@@ -195,14 +177,15 @@ export default function FilterModal({
               Your ID List results in:
             </b>
             {/* section of matched and unmatched ID */}
-            <MatchResults numMatched={numMatched} numUnmatched={numUnmatched} />
+            <MatchResults numMatched={numMatched} />
```
daniel-cy-lu, line 180: remove all codes associated with the unmatched id field.

components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/index.tsx
```diff
@@ -89,48 +88,31 @@ export default function FilterModal({
...
       const submitterIdMatch = filterSubmitterIds.includes(result.submitterDonorId);
 
-      if (donorIdMatch || submitterIdMatch) {
+      if (donorIdMatch) {
         filteredTextAreaIDs.add(result.donorId);
       }
+      if (submitterIdMatch) {
+        filteredTextAreaIDs.add(result.submitterDonorId);
+      }
     });
```
daniel-cy-lu, line 110: Update the logic of the matched id count so it counts donor and submitter id from the same entry as 2.

demariadaniel:
One issue I've found while testing: if you enter a single Submitter ID into the Modal, you will see the results are `DONaN` and the filter is not working. Compared with Develop and the issue seems to be related to this branch. Can you check into this?

daniel-cy-lu:
> One issue I've found while testing: if you enter a single Submitter ID into the Modal, you will see the results are `DONaN` and the filter is not working. Compared with Develop and the issue seems to be related to this branch. Can you check into this?
> 
> Screen Shot 2023-03-29 at 3 31 51 PM
> 
> Screen Shot 2023-03-29 at 3 31 32 PM

@demariadaniel Dan, I found the cause. I can only add donor id to the set or it will change the down stream codes. So that's what I did. And I added a counter to keep track of the matched id count.

demariadaniel:
> > One issue I've found while testing: if you enter a single Submitter ID into the Modal, you will see the results are `DONaN` and the filter is not working. Compared with Develop and the issue seems to be related to this branch. Can you check into this?
> > Screen Shot 2023-03-29 at 3 31 51 PM
> > Screen Shot 2023-03-29 at 3 31 32 PM
> 
> @demariadaniel Dan, I found the cause. I can only add donor id to the set or it will change the down stream codes. So that's what I did. I can't use the set.size to keep track, since that forces me to add submitter id to the set and I don't want to do that. So I just keep count with `matchedCount`.
> 
> Screenshot 2023-03-29 at 4 51 33 PM

Looks like it's working as expected on localhost, can you update the branch w/ develop? Shouldn't be any conflicts but let me know if there are

components/pages/submission-system/program-submitted-data/SearchBar/FilterModal/index.tsx
daniel-cy-lu: This logic will not add 1 to the count when users enter the donor id and submitter id from the same entry. I was thinking your way initially. But had to change it to the three if statements. Would be nice to refactor it, but I ran out of idea haha

demariadaniel: Oh I get it now. Sorry I misread how this works, because we're counting these as 2 separate matches.

daniel-cy-lu:
> Looks like it's working as expected on localhost, can you update the branch w/ develop? Shouldn't be any conflicts but let me know if there are

No merge conflict in local branches when doing a merge with up to date develop.

demariadaniel (approved):

Issue #2568, opened 2023-02-08
Update File Repository Table


samrichca:
Update File Repository Table

PR #2646, opened 2023-04-24
#2568 - update File Repository table to React Table 8

samrichca:
# Description of changes

- update the table on the File Repository page to React Table 8
- requires uikit branch https://github.com/icgc-argo/uikit/pull/116

## Type of Change

- [ ] Bug
- [x] Dependency updates
- [ ] Feature
- [x] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [x] No changes to UI
- Back-end (select one):
  - [x] No back-end changes
- [x] Manually tested changes
- [x] Added copyrights to new files
- [x] Connected ticket to PR


components/pages/file-repository/FileTable/index.tsx
```diff
@@ -314,13 +322,23 @@ const FileTable = () => {
...
+    ...column,
+    meta: { customCell: true },
+    ...(column.cell
+      ? {}
+      : {
+          cell: ({ row: { original }, cell }) => (
+            <RowSelectionCellWrapper original={original}>{cell.getValue()}</RowSelectionCellWrapper>
+          ),
+        }),
+  }));
```
ciaranschutte, line 339: could this be cleaner? 

The double mapping I don't understand the need for. Maybe it's because you want `meta: { customCell: true },`
to be in one place that's easy to change, instead of every single cell. Then you could have a function that adds the meta in the cell if applicable. something like:
```javascript
...
size: 60,
meta: addMeta({customCell: true});
...
```

Then the second part might make more sense to just have a `cell` property and conditionally have it as empty object or the function. currently it's spreading just a `cell` element or nothing ( {} ). this way it also means that there will aways be a `cell` property which helps for typing. I'm working through a lot of TS knowledge rn and it's helped me to try and always have a property with a defined type, without conditionally adding properties in this manner. helps avoid generics too.

ciaranschutte (approved):
suggestion not request for changes

components/pages/file-repository/FileTable/index.tsx
```diff
@@ -96,206 +104,206 @@ const useFileRepoTableQuery = (
...
-      Header: 'File ID',
-      headerClassName: pagingState.sort == DEFAULT_SORT ? '-sort-desc' : '',
-      id: FileCentricDocumentField.file_id,
-      accessor: 'fileId',
-      Cell: ({ original }: { original: FileRepositoryRecord }) => {
+      header: () => (
+        <TableRowSelectionCheckbox
+          checked={allRowsSelected}
+          id="toggleAll"
+          onChange={toggleAllHandler}
```
ciaranschutte, line 181: interesting this doesn't throw a TS error  because 
```javascript
  const toggleAllHandler = () => {
    setSelectedRowIds([]);
    setUnselectedRowIds([]);
    setAllRowsSelected(!allRowsSelected);
  };
```

doesnt match this signature: 
```javascript
export const TableRowSelectionCheckbox = ({
  checked,
  id,
  onChange,
}: {
  checked: boolean;
  id: string;
  onChange: (id: string) => void;
}) => (
  <Checkbox
    aria-label="table-row-select"
    checked={checked}
    onChange={() => onChange(id)}
    value={id}
  />
);
```

samrichca: I think it's because toggleAllHandler isn't asking for an arg but TableRowSelectionCheckbox onChange is providing one. if it were the other way around there'd be an error

components/pages/file-repository/FileTable/index.tsx
samrichca: I've added a default column and default cell to improve this

I posted this on slack and I'll add it to the readme... the type for the columns array is ColumnDef<DataShape> which is from the library and it's a partial. we should only add the properties we're overriding from the defaults provided by the library. this feedback is good general advice though :+1:

Issue #2679, opened 2023-03-14
Build The New Submit Clinical Page In Platform 


Buwujiu:
## Detailed Description
- Add a new Submit Clinical Data page in argo platform, this will eventually replace the current clinical submission pages. 
- User story:
As a user, I want to access the RDPC portal from ARGO Platform, so that I can submit clinical data. 
Refer to 5. User Story - Submit Clinical Data Page in https://wiki.oicr.on.ca/display/icgcargotech/Program+And+User+Management+System

- Mockup: 
- https://www.figma.com/file/O9jdbXSZMQf5F34OJAhyRT/ARGO-RPDC%3A-Exploration?type=design&node-id=446-109691&t=pNL7asaj8v2B0WgM-4
![image](https://github.com/icgc-argo/roadmap/assets/8744047/65872a62-d228-4b04-84c2-31d8ca2ef85d)

## Exit Criteria
1. When clicking on the link "Submit Data: Access RDPC Portal to submit clinical and molecular data.", users should be directed to the assigned RDPC portal (there is only one right now). TODO: insert the rdpc portal link. 
2. The Program Summary table should show the basic public information of the program. 
3. **Since we are not removing existing clinical submission pages, can we add a feature flag to enable the new page in QA, hide it in staging and prod?**

Buwujiu:
Found issues and logged in https://github.com/icgc-argo/roadmap/issues/1002.

PR #2684, opened 2023-06-26
📄 #2679 Submit Clinical Page

demariadaniel:
# Description of changes

Basic Page w/ RDPC Banner Link + Summary Table
w/ Feature Flagging & Program links

[UI #2679](https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/platform-ui/2679)

## Type of Change

- [ ] Bug
- [ ] Dependency updates
- [x] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [ ] No back-end changes
- [ ] Manually tested changes
- [ ] Connected ticket to PR


pages/submission/program/[shortName]/submit-clinical.tsx
```diff
@@ -0,0 +1,43 @@
...
+      canWriteProgramData({ permissions, programId: String(shortName) })
+    );
+  },
+  startWithGlobalLoader: true,
+})((props) => {
+  const { FEATURE_SUBMIT_CLINICAL_ENABLED } = getConfig();
+
+  if (!FEATURE_SUBMIT_CLINICAL_ENABLED) return <ErrorPage statusCode={404} />;
+  return <ProgramSubmitClinical {...props} />;
+});
```
Comment (ciaranschutte) [inline pages/submission/program/[shortName]/submit-clinical.tsx:43]: I prefer this style of page error: https://github.com/icgc-argo/platform-ui/blob/develop/pages/program/%5BprogramId%5D.tsx 

For consistency + helps if there's any additional handling ever implemented or ErrorPage changes

global/constants/argoPages.tsx
```diff
@@ -22,3 +22,5 @@ export const ARGO_ROOT = 'https://www.icgc-argo.org';
 export const ARGO_PRIVACY_PAGE = `${ARGO_ROOT}/page/2/privacy`;
 export const ARGO_TERMS_PAGE = `${ARGO_ROOT}/page/1/terms-and-conditions`;
 export const ARGO_PUBLICATION_PAGE = `${ARGO_ROOT}/page/77/e3-publication-policy`;
+
+export const RDPC_PORTAL_URL = `http://localhost:3000`;
```
ciaranschutte, line 26: good use case for env var - will be able to update this in deployment config instead of hard coding in another code change

ciaranschutte: can default to localhost

components/pages/submission-system/SideMenu.tsx
```diff
@@ -276,6 +277,23 @@ const LinksToProgram = (props: { program: SideMenuProgram; isCurrentlyViewed: bo
               />
             </NextLink>
           )}
+          {FEATURE_SUBMIT_CLINICAL_ENABLED && (
+            <NextLink
+              as={`${PROGRAM_SUBMIT_CLINICAL_PATH.replace(
+                PROGRAM_SHORT_NAME_PATH,
+                props.program.shortName,
+              )}`}
```
ciaranschutte, line 285: this is a string already, doesn't need to be in a template literal

ciaranschutte:
if this will be behind a feature flag it's worthwhile to put it into `.env.schema` FEATURE_SUBMIT_CLINICAL_ENABLED
(one thing to consider is we need up with a lot of feature flags that are always TRUE)

ciaranschutte:
copyrights on newly created pages are 2022 instead of 2023

ciaranschutte:
Screenshot 2023-06-28 at 10 32 40 PM

gutters in the main section look wider in consistency with other pages

ciaranschutte:
double side menu item for submit clinical date 
sidemnu

unclear if this is intentional due to a need to keep old page alongside new page - either way it's one that may easily slip through. possibly add a conditional in the side menu - a ternary for the flag ? `new_page_enabled ? <NewPage> : <OldPage>`?

ciaranschutte:
Screenshot 2023-06-28 at 11 09 53 PM
Screenshot 2023-06-28 at 11 10 31 PM

Just my two cents but I think it looks better if we change the loading state to the regular loader.
Something like: 
```typescript
        {loading ? (
                <div
                  css={css`
                    display: flex;
                    justify-content: center;
                    align-items: center;
                    width: 100%;
                  `}
                >
                  <DnaLoader />
                </div>
              ) : (
                <ProgramSummaryTable
                  data={
                    loading ? { ...programSummaryData, Website: 'loading' } : programSummaryData
                  }
                  title={'Program Summary'}
                />
              )}
              }
```

There might be a centered loader already but I couldn't find it with a quick look. The "global loader" is intended for app level loading I think rather than section loading

ciaranschutte:
[ICGC ARGO _ ALEXIS-CA Dashboard.webm](https://github.com/icgc-argo/platform-ui/assets/1486054/b6c7082a-924e-48a6-ad8c-09bf58895db1)

There's some odd hovering states on this `<SimpleTable>` component which I can't recreate in UIKit.
The grey rows go white when hovered - which doesn't make any sense because it should be display only.
Bandaid (not a fully thought through solution) might be to pass additional css for hover states to override.

pages/submission/program/[shortName]/submit-clinical.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default createPage({
+  isPublic: false,
+  isAccessible: async ({ ctx, initialPermissions: permissions }) => {
+    const {
+      query: { shortName },
+    } = ctx;
+    return (
+      canReadProgram({ permissions, programId: String(shortName) }) &&
+      canWriteProgramData({ permissions, programId: String(shortName) })
```
Comment (ciaranschutte) [inline pages/submission/program/[shortName]/submit-clinical.tsx:34]: I know this `String()` way of doing is used elsewhere but to me it's just coding around TS rather than using it.

Quick solution there's two options: 
1) use an assertion - but make sure it's a valid case - likely something one of the seniors assigned review can chime in on
2) make sure it's a string and not a string array - we're accepting that TS is warning us about potential issues and making sure we account for them.

something like: 
```typescript
    const shortName = Array.isArray(shortName) && shortName.length ? shortName[0] : shortName;
    // ^ this makes sense to me but TS is giving me an error, so might need a tinker.
    const shortName = Array.isArray(shortName) ? shortName[0] : shortName;
    // ^ this one narrows correctly but I'm unclear why it's not calling out `shortName[0]` maybe being undefined
```

Longer solution:
Extend the root page `isAccessible` function to do additional parsing and/or extend types which we're already augmenting with declaration merging here: https://github.com/icgc-argo/platform-ui/blob/develop/global/utils/pages/types.tsx#L27

something like: 
```typescript
    page.isAccessible = async () => isAccessible ? isAccessible() :  true);

```
here: https://github.com/icgc-argo/platform-ui/blob/develop/global/utils/pages/index.tsx#L74

ciaranschutte:
@demariadaniel looks great - I know i'm not an assigned reviewer so take it all as comments :)

global/constants/argoPages.tsx
demariadaniel: Yes will also be multiple env, updated

components/pages/submission-system/SideMenu.tsx
demariadaniel: Right, updated

demariadaniel:
> if this will be behind a feature flag it's worthwhile to put it into `.env.schema` FEATURE_SUBMIT_CLINICAL_ENABLED so people know exactly what to put in deployment vs looking through code (one thing to consider is we end up with a lot of feature flags that are always TRUE)

Yes, added!

> copyrights on newly created pages are 2022 instead of 2023

Thanks fixed

pages/submission/program/[shortName]/submit-clinical.tsx
Comment (demariadaniel) [reply pages/submission/program/[shortName]/submit-clinical.tsx:38]: Sounds good! Updated

pages/submission/program/[shortName]/submit-clinical.tsx
Comment (demariadaniel) [reply pages/submission/program/[shortName]/submit-clinical.tsx]: OK I've updated the context query prop type in `/global/utils/pages`, seems best to address the issue at the root, unless there are cases where `ctx.query.shortName` will in fact be a string[]

demariadaniel:
> ICGC.ARGO._.ALEXIS-CA.Dashboard.webm 
> There's some odd hovering states on this `<SimpleTable>` component which I can't recreate in UIKit. The grey rows go white when hovered - which doesn't make any sense because it should be display only. Bandaid (not a fully thought through solution) might be to pass additional css for hover states to override.

Discussed in Slack - this behaviour is showing on File / Donor / Program Entity pages as well, deserves a separate ticket

ciaranschutte (approved):

PR #2732, opened 2023-10-30
Feat/clinical data downloads

joneubank:
# Description of changes

Adds ability to download clinical data for selected files through the UI. Clinical Data download is available in the download actions dropdown of the File Repository and a dropdown added to the File Details page.

A feature flag has been added to the environment config, in order to activate clinical downloads the env variable `FEATURE_CLINICAL_DOWNLOAD=true` must be set.

There are a couple deviations from the design that can be corrected through updates to the uikit:
1. download dropdown options don't get disabled when unavailable, instead they are removed from the menu.
2. the table icon by the file page's Legend button has not been removed

Additionally, the number of donors selected is not displayed, there will need to be a slightly complicated modification to the UI to accomplish this.

## Type of Change

- [ ] Bug
- [ ] Dependency updates
- [x] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [x] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [x] Back-end PRs merged, tested on develop, and linked in this PR
  - [ ] No back-end changes
- [x] Manually tested changes
- [ ] Connected ticket to PR


components/pages/file-repository/FacetPanel/types.ts
```diff
@@ -36,7 +38,7 @@ export type SearchMenuDataNode = {
 export type FacetDetails = {
   name: string;
   facetPath: FileFacetPath;
-  variant: 'Basic' | 'Number' | 'Tooltip' | 'Other';
+  variant: 'Basic' | 'Number' | 'Tooltip' | 'ClinicalData' | 'Other';
```
joneubank, line 41: Created a variant just to pass in a handler for the custom display values (converting true to 'Available', etc.)... I'm not sure if having dedicated variant for fields with display name considerations is best or if finding a way to modify the display text info is the right way, but this was simple...

components/pages/file-repository/FileTable/TsvDownloadButton.tsx
```diff
@@ -77,11 +94,14 @@ const TsvDownloadButton = ({
...
-    //   value: DownloadOptionValues.CLINICAL_DATA,
-    // },
+    ...(!!selectedFilesCount && FEATURE_CLINICAL_DOWNLOAD
+      ? [
+          {
+            display: 'Clinical Data',
+            value: DownloadOptionValues.CLINICAL_DATA,
+          },
+        ]
+      : []),
```
joneubank, line 104: only show "clinical data" download option when files have been selected. Don't have a good way at this moment to "download all" clinical data by filter. Could add that through the UI though like we have for file data.

components/pages/file-entity/index.tsx
```diff
@@ -51,21 +49,22 @@ const FileEntity = ({ fileId }) => {
...
     fileId,
   });
-  const { egoJwt } = useAuthContext();
-  const { data: userProfile, loading: profileLoading } = useQuery(USER_PROFILE_QUERY);
+  const { egoJwt, permissions } = useAuthContext();
+
   const { MAX_FILE_DOWNLOAD_SIZE } = getConfig();
 
-  const loading = profileLoading || fileLoading;
+  const loading = fileLoading;
```
justincorrigible, line 56: what makes this necessary? 
line 48 could just be renamed, or use fileLoading in 76

components/pages/file-repository/FacetPanel/index.tsx
```diff
@@ -762,18 +777,19 @@ const FacetPanel = () => {
           presetFacets
             // Conditionally filter embargo_stage facet
             .filter((f) => {
-              return embargoStageEnabled || f.facetPath !== FileFacetPath.embargo_stage;
+              return f.facetPath === FileFacetPath.embargo_stage ? embargoStageEnabled : true;
```
justincorrigible, line 780: please forgive my being dense, what is `f` again? could that variable be renamed to something a bit more declarative?

components/pages/file-entity/DownloadDropdown.tsx
```diff
@@ -0,0 +1,186 @@
...
+          .then(() => setLoading(false))
+          .catch((err) => {
+            setLoading(false);
+            toaster.onDownloadError(err.message);
+          });
+        break;
+      }
+      case DropdownOptions.FILE: {
+        const downloadUrl = urljoin(GATEWAY_API_ROOT, API_PATH_DOWNLOAD_FILE, props.objectId);
+        // window.location.assign(tsvdownloadUrl);
```
demariadaniel, line 90: This can be removed?

components/pages/file-repository/FileTable/TsvDownloadButton.tsx
```diff
@@ -49,11 +63,14 @@ const TsvDownloadButton = ({
 }: {
   allFilesSelected: boolean;
   selectedFilesCount: number;
+  // selectedDonorsCount: number;
+  // selectedFilesWithClinicalData: string[];
```
justincorrigible, line 67: these could be removed?

components/pages/file-repository/utils/index.ts
```diff
@@ -227,11 +227,40 @@ export const removeFilter: TRemoveFilter = (field, filters) => {
     : defaultEmptyFilters;
 };
 
-export const toDisplayValue: (value: string, field?: string) => string = (value, field) => {
+export const clinicalDisplayValues = (value: string) => {
+  return value === 'false'
```
justincorrigible, line 231: it may be worth accounting for casing. just to be safe.

components/pages/file-repository/FileTable/TsvDownloadButton.tsx
```diff
@@ -127,22 +147,49 @@ const TsvDownloadButton = ({
...
           )}&columns=${encodeURIComponent(JSON.stringify(fileRepoTableTSVColumns))}`,
         );
         // window.location.assign(tsvdownloadUrl);
-        downloadFileWithEgoToken(tsvdownloadUrl).finally(() => setLoading(false));
+        downloadFileWithEgoToken(tsvdownloadUrl)
+          .then(() => setLoading(false))
+          .catch((err) => {
+            setLoading(false);
+            toaster.onDownloadError(err.message);
+          });
```
joneubank, line 174: This is the working way to handle creating toasters on error while still updating the loading state. Having state update in both a catch block and the finally block would cause state update on unmounted components, so finally doesn't work in that case. Maybe you have a better idea how to handle this? try/catch could be cleaner maybe

components/pages/file-repository/types.ts
```diff
@@ -1,25 +1,30 @@
...
+  file_id: 'file_id',
+  file_number: 'file_number',
+  file_type: 'file_type',
+  has_clinical_data: 'has_clinical_data',
+  object_id: 'object_id',
+  release_state: 'release_state',
+  study_id: 'study_id',
+};
+
+export type FileCentricDocumentField = Values<typeof FileCentricDocumentFields>;
```
joneubank, line 27: Remove enum, replace with const plus union type.

generated/gql_types.tsx
```diff
@@ -2319,6 +2320,7 @@ export type FileNode = Node & {
   file_id?: Maybe<Scalars['String']>;
   file_number?: Maybe<Scalars['Float']>;
   file_type?: Maybe<Scalars['String']>;
+  has_clinical_data?: Maybe<Scalars['Boolean']>;
```
justincorrigible, line 2323: fun! I don't recall seeing `Maybe`. will do some reading

global/hooks/toaster.tsx
```diff
@@ -40,8 +40,6 @@ export const useToastState = () => {
   const [toastStack, setToastStack] = useState<(ToastConfig & { id: string })[]>([]);
 
   const addToast = (toast: ToastConfig & { timeout?: number }) => {
-    console.log(`🔥🍞🍞🍞🍞🍞🍞🔥`);
-    console.log(`🔥🔥🔥🔥🔥🔥🔥🔥`);
```
justincorrigible, line 44: 😆

components/pages/file-repository/utils/index.ts
```diff
@@ -227,11 +227,40 @@ export const removeFilter: TRemoveFilter = (field, filters) => {
...
+  // Custom display values for `has_clinical_data` boolean field
+  if (field === 'has_clinical_data') {
+    return value === 'false'
+      ? 'Not Available'
+      : value === 'true'
+      ? 'Available'
+      : value === IS_MISSING
+      ? 'No Data'
+      : value;
+  }
```
joneubank, line 254: This function is used by the SQON view component, so it also has a copy of the clinicalDisplayValue conversion.

I will update this to reference the other function for DRY reasons.

components/pages/file-repository/utils/index.ts
```diff
@@ -227,11 +227,40 @@ export const removeFilter: TRemoveFilter = (field, filters) => {
...
+
+  // Display values for Boolean fields
+  if (value === 'false') {
+    return 'False';
+  }
+  if (value === 'true') {
+    return 'True';
+  }
```
joneubank, line 262: Not currently used, but useful for future cases that have boolean values?

justincorrigible (approved):
minor improvements suggested. 
code looks good to squash and merge as is, regardless.

components/useCommonToasters.tsx
```diff
@@ -23,6 +23,20 @@ import { useToaster } from 'global/hooks/toaster';
 export default function useCommonToasters() {
   const toaster = useToaster();
   return {
+    onDownloadError: (error?: string) => {
+      toaster.addToast({
+        variant: TOAST_VARIANTS.ERROR,
+        title: `Download Error`,
+        content: error || 'An error occurred, the file could not be downloaded.',
+      });
+    },
```
joneubank, line 32: Consistent toasters for failed downloads.

global/hooks/toaster.tsx
joneubank, line 44: this was fun but not ever meant for production.

global/hooks/useAuthContext.tsx
```diff
@@ -121,11 +121,20 @@ export function AuthProvider({
...
     const response = await fetchWithEgoToken(uri, options);
+
+    if (!response.ok) {
+      const data = await response.json().catch((_) => {
+        console.log(`Download request failed and returned non-json response.`);
+      });
+      throw new Error(
+        data?.error || 'Something went wrong with the attempted download. Please try again later.',
+      );
+    }
```
joneubank, line 137: Without this handler, download requests that had a server error would cause the error message to be downloaded as a file.

global/hooks/toaster.tsx
justincorrigible: totally agreed. unless behind a dev mode flag

global/utils/egoJwt/index.tsx
```diff
@@ -26,7 +26,7 @@ const TokenUtils = createEgoUtils(getConfig().EGO_PUBLIC_KEY);
 
 type PermissionScopeObj = {
   policy: string;
-  permission: 'READ' | 'WRITE' | 'ADMIN' | 'DENY';
+  permission: 'READ' | 'WRITE' | 'DENY';
 };
```
joneubank, line 30: Ideally this type would be exported from the ego utils instead of being redefined here.

generated/gql_types.tsx
joneubank: Its definied locally in this file, not a TS default. Useful in exactly this GQL return type use case, but generally we don't want to be introducing null types in TS code.

components/pages/file-repository/utils/index.ts
joneubank: generally a good idea. ES boolean field string values will be exactly `false` and `true`, any other value would be interpreting something unexpected...

this would protect against changes in the return type from arranger in the future?

generated/gql_types.tsx
justincorrigible: `Maybe` just seems useful when you're dealing with clinical data these days. lots of actual `null` values here and there to account for, so this is great as a pattern!

PR #2742, opened 2023-11-16
improve recent announcements

justincorrigible:
bonus, re-syncs outdated:
- manifest lock
- graphql type definitions

### DIFF
--- a/components/pages/Homepage/common.tsx
+++ b/components/pages/Homepage/common.tsx
@@ -100,28 +100,14 @@ export const DataReleaseBar: ComponentType<{
   ) : (
     <div
       css={css`
-        margin: 20px 0px 40px;
+        border-top: 4px solid ${theme.colors.accent3_2};
+        margin: 40px 0px;
         width: 80%;
       `}
     >
-      <div
-        css={css`
-          margin-bottom: 5px;
-        `}
-      >
-        <Typography color={theme.colors.white} variant={'data'}>
-          Data Release {version.releaseIteration} - {version.date}
-        </Typography>
-      </div>
-      <div
-        css={css`
-          border-top: 4px solid ${theme.colors.accent3_2};
-        `}
-      >
-        <Row nogutter justify={'center'}>
-          {formattedStats}
-        </Row>
-      </div>
+      <Row nogutter justify={'center'}>
+        {formattedStats}
+      </Row>
     </div>
   );
 };
@@ -296,16 +282,6 @@ export const NewsContainer: ComponentType<{ newsItems: NewsItem[] }> = ({ newsIt
                 border-bottom: 1px solid ${theme.colors.grey_2};
               `}
             >
-              <Typography
-                bold
-                component="h3"
-                variant="paragraph"
-                css={css`
-                  margin: 0;
-                `}
-              >
-                Announcements:
-              </Typography>
               {newsItems.map((newsItem: NewsItem, index: number) => (
                 <Typography
                   key={`newsItem-${index}`}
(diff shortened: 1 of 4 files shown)


components/pages/Homepage/common.tsx
```diff
@@ -100,28 +100,14 @@ export const DataReleaseBar: ComponentType<{
...
         width: 80%;
       `}
     >
-      <div
-        css={css`
-          margin-bottom: 5px;
-        `}
-      >
-        <Typography color={theme.colors.white} variant={'data'}>
-          Data Release {version.releaseIteration} - {version.date}
```
justincorrigible, line 113: removes this pesky little thing nobody cares about

components/pages/Homepage/common.tsx
```diff
@@ -100,28 +100,14 @@ export const DataReleaseBar: ComponentType<{
   ) : (
     <div
       css={css`
-        margin: 20px 0px 40px;
+        border-top: 4px solid ${theme.colors.accent3_2};
```
justincorrigible, line 103: while keeping the formatting/space

components/pages/Homepage/index.tsx
```diff
@@ -67,6 +67,24 @@ const newsItems: NewsItem[] = [
...
+        >
+          Data Release 7.0
+        </Link>
+        , and{' '}
+        <Link
+          href="https://docs.icgc-argo.org/docs/release-notes/software-releases#november-6-2023-clinical-downloads"
+          rel="noopener noreferrer"
+          target="_blank"
+        >
+          Software Release UI 1.125.1 - API 3.42.0
```
justincorrigible, line 85: adds release info to the nov 6 announcement

components/pages/Homepage/index.tsx
```diff
@@ -105,50 +123,6 @@ const newsItems: NewsItem[] = [
       </>
     ),
   },
-  {
-    title: 'June 8, 2023',
```
justincorrigible, line 109: removing the noise of that fourth announcement... we're supposed to only have 3 at a time

components/pages/Homepage/index.tsx
```diff
@@ -105,50 +123,6 @@ const newsItems: NewsItem[] = [
...
-        <br />
-        <br />
-        Another new feature is also available on our platform: The Program Entity Page. Accessible
-        from the File Repository, this page presents a summary of the programs, offering a
-        high-level overview of donor and file statistics.
-      </>
-    ),
-  },
-  {
-    title: '',
```
justincorrigible, line 125: and gets rid of this random thing that was adding implementation confusion

generated/gql_types.tsx
```diff
@@ -401,6 +401,23 @@ export type CoreCompletionStatusCount = {
   noData: Scalars['Int'];
 };
 
+export type DataCenter = {
```
justincorrigible, line 404: looks like the datacenter stuff was added, but the types weren't updated

package-lock.json
justincorrigible, line 1: and this file had fallen out of sync, so we couldn't do `npm ci`, as we should

generated/gql_types.tsx
ciaranschutte: nice! just a heads up .. there's only two places were the gql generated types are actually being used. the rest of the project gql types were written by us. they are meant to match with gql return types. these types are not the same in many ways, not least that they need to be manually updated.

the gql codegen commands + libs were added but changes weren't made to use the generated types instead.

Came across this because I based the gql typing for `rdpc-ui` off this project. Wondered why I was banging so many walls but everything was gravy in `platform-ui`

ciaranschutte (approved):

generated/gql_types.tsx
justincorrigible: ah! so no need to update this file?
lol why commit it to begin with, then? 🤔😂

I'll remove it some other time.
and thanks for the explainer!

PR #2763, opened 2024-04-11
↕️ Submitted Data: Sort Invalid Donors First

demariadaniel:
# Description of changes

https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/platform-ui/2624

Sorts Invalid Records to the top of the Submitted Data table on first page load
examples shown in ticket

## Type of Change

- [x] Bug
- [ ] Dependency updates
- [ ] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [ ] No back-end changes
- [ ] Manually tested changes
- [ ] Connected ticket to PR

### DIFF
--- a/components/pages/submission-system/program-submitted-data/ClinicalEntityDataTable/index.tsx
+++ b/components/pages/submission-system/program-submitted-data/ClinicalEntityDataTable/index.tsx
@@ -142,12 +142,18 @@ const getColumnWidth = memoize<
 const defaultEntityPageSettings = {
   page: defaultClinicalEntityFilters.page,
   pageSize: defaultClinicalEntityFilters.pageSize,
-  sorted: [{ id: 'donorId', desc: true }],
+  sorted: [
+    { id: 'schemaMetadata.isValid', desc: false },
+    { id: 'donorId', desc: true },
+  ],
 };
 
 const defaultDonorSettings = {
   ...defaultEntityPageSettings,
-  sorted: [{ id: 'completionStats.coreCompletionPercentage', desc: false }],
+  sorted: [
+    { id: 'schemaMetadata.isValid', desc: false },
+    { id: 'completionStats.coreCompletionPercentage', desc: false },
+  ],
 };
 
 const defaultErrorPageSettings = {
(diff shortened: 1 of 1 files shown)


components/pages/submission-system/program-submitted-data/ClinicalEntityDataTable/index.tsx
joneubank, line 1: These sorting configs will only organize the records sent to the table, which is one page at a time. So if you have records that are invalid those will now appear at the top of the table, but records on subsequent pages will not be pulled onto the first page with this. To do that would require sorting of the data on the server-side.

joneubank (dismissed):
No problem with these changes, but I am not sure they fulfill the intention of the ticket, since there could still be records with errors that will not show on first load since they are found on later pages.

components/pages/submission-system/program-submitted-data/ClinicalEntityDataTable/index.tsx
demariadaniel: You are correct, and this definitely shows there is something fishy going on

There are reasons I went with this route:
a) The sorting configs also determine the `sort` argument passed to our query, see: https://github.com/icgc-argo/platform-ui/blob/develop/components/pages/submission-system/program-submitted-data/ClinicalEntityDataTable/index.tsx#L222

To me this was logical; keep the API as is, returning all records by DonorId as the standard, and have the UI Table request those records specifically arranged by `isValid`


b) If you look at my example screenshots in the ticket, it would _appear_ that this was sorting records from subsequent pages to the top. There were 0 errors on the DATA-CA Specimens page 1 in Develop, then in this branch you see 4 errors for the same page.

However, when you expand the page size to 100 records, the total number of errors increases to 6, so pagination is still affecting sorting. And, while there are less than 20 Donors w/ errors, not all of those Donors are included in those first 20 records returned.
So the updated query is not doing the whole job of prioritizing invalid records.


So I will work on a Clinical solution to insure all invalid records come first
And I'll investigate why the DATA-CA Specimen page is not currently showing any errors on page 1 b/c there are errors related to these first few donors

joneubank: Probably this fix and one to clinical are both required. Thanks for reviewing!

demariadaniel:
Related Clinical changes here:
https://github.com/icgc-argo/argo-clinical/pull/1184

joneubank (approved):

PR #2795, opened 2025-02-05
Feat/charts bar

ciaranschutte:
# Description of changes

Add charts page skeleton and bar charts config
Theming is dependant on React lifecycle because we use a hook to inject themes
`api` is components we can extract to Arranger if we decide to

## Type of Change

- [ ] Bug
- [ ] Dependency updates
- [ ] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [ ] No back-end changes
- [ ] Manually tested changes
- [ ] Connected ticket to PR

### DIFF
--- a/components/pages/discovery/Charts.tsx
+++ b/components/pages/discovery/Charts.tsx
@@ -17,10 +17,7 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { css, PageContainer, styled, UikitTheme, useTheme } from '@icgc-argo/uikit';
-import NavBar from 'components/NavBar';
-import { Row, setConfiguration } from 'react-grid-system';
-import Head from '../head';
+import { css, UikitTheme, useTheme } from '@icgc-argo/uikit';
 
 import Card from './components/Card';
 import { donorData } from './data';
@@ -127,6 +124,7 @@ const Charts = () => {
       left: 24,
       bottom: 56,
     },
+
     colorBy: 'indexValue',
     theme: chartTheme.theme,
   });
--- a/components/pages/discovery/Charts.tsx
+++ b/components/pages/discovery/Charts.tsx
@@ -1,0 +1,188 @@
+/*
+ * Copyright (c) 2025 The Ontario Institute for Cancer Research. All rights reserved
+ *
+ * This program and the accompanying materials are made available under the terms of
+ * the GNU Affero General Public License v3.0. You should have received a copy of the
+ * GNU Affero General Public License along with this program.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { css, PageContainer, styled, UikitTheme, useTheme } from '@icgc-argo/uikit';
+import NavBar from 'components/NavBar';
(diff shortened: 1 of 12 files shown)


components/pages/discovery/Charts.tsx
```diff
@@ -0,0 +1,188 @@
...
+      domain: {
+        line: {
+          stroke: injectedTheme.colors.grey_2,
+          strokeWidth: 1,
+        },
+      },
+    },
+  },
+});
+
```
ciaranschutte, line 63: nivo inline style svg elements so it's best to make a nivo theme from project theme and pass the entire object.
otherwise you're fighting against inline styles , sprinkling "!important" for svg elements

components/pages/discovery/Charts.tsx
```diff
@@ -0,0 +1,188 @@
...
+  '#FEA430',
+  '#9BC7ED',
+  '#A1A4B1',
+  '#CBBBEA',
+  '#FA8564',
+  '#B7EBF4',
+  '#A7F0E1',
+  '#FCBEAC',
+  '#FEE8CB',
+];
```
ciaranschutte, line 81: Some are Argo theme colours. Some are not.
There's a huge mixture of theme usage throughout this project that it's broken.
There are inline colours in many places already which means inconsistency.
The theming in general needs a reviews as I don't think it makes sense.
eg. why would I use `warning_1` for a chart bar colour that has nothing to do with a warning? Unfortunately we've loaded the theme system with meaning.

components/pages/discovery/Charts.tsx
```diff
@@ -0,0 +1,188 @@
...
+        <BarChart data={donorData} config={programIDBarChart} />
+      </Card>
+
+      <Card title="RDPC Node" css={css({ gridColumnStart: 2, gridRowEnd: 'span 1' })}>
+        <BarChart data={donorData} config={programIDBarChart} />
+      </Card>
+
+      <Card title="Track Embargo State" css={css({ gridColumnStart: 2, gridRowEnd: 'span 1' })}>
+        <BarChart data={donorData} config={programIDBarChart} />
+      </Card>
```
ciaranschutte, line 146: CSS grid is the most practical, easy to tweak, not build ourselves into a corner way of doing this kind of layout.
We don't need a fancy component at this time. CSS grid is supported everywhere and easy to use.

components/pages/discovery/components/api/BarChart.tsx
```diff
@@ -0,0 +1,115 @@
...
+ * Create bar chart config from consumer provided config
+ *
+ * @param config - See Nivo bar chart config docs - must supply 'data', 'indexBy' and 'keys'
+ * @returns A Nivo bar chart configuration object
+ */
+export function createBarConfig(consumerConfig: RequiredConfig): RequiredConfig {
+  const config = { ...baseConfig, ...consumerConfig };
+  return config;
+}
+
```
ciaranschutte, line 85: intentionally light for now

components/pages/discovery/components/api/BarChart.tsx
```diff
@@ -0,0 +1,115 @@
...
+ * Defaults to response dimensions
+ *
+ * @returns BarChart component
+ */
+const BarChart = ({
+  data,
+  config,
+}: {
+  data: Record<string, string | number>[];
+  config: RequiredConfig | ((baseConfig: BarChartConfig) => RequiredConfig);
```
ciaranschutte, line 104: config similar to how webpack configurations work, if consumer needs access to the defaults they can supply a function instead of an object

components/pages/discovery/Charts.tsx
```diff
@@ -0,0 +1,188 @@
...
+    margin: {
+      top: 12,
+      right: 24,
+      left: 24,
+      bottom: 56,
+    },
+    colorBy: 'indexValue',
+    theme: chartTheme.theme,
+  });
+
```
ciaranschutte, line 133: not ideal to generate this in the component but it's a very light object and our hands are tied by our usage of theming through react hooks. we can optimise later if the need arises

components/pages/discovery/components/api/Chart.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { css } from '@emotion/react';
+import { ReactNode } from 'react';
+
+export const Chart = ({ children, height }: { children: ReactNode; height?: any }) => (
```
justincorrigible, line 23: I recall there's a `WithChildren` props type that may be more maintainable (as it comes directly from @types/react, and gets updated along with the library). may want to look into that instead of explicitly defining the type for children

components/pages/discovery/components/api/Chart.tsx
```diff
@@ -0,0 +1,27 @@
...
+  <div className="scroll" css={css({ flex: 1, overflow: 'scroll' })}>
```
justincorrigible, line 24: to make components more reusable —while not strictly necessary— I tend to use cx/classnames to interpolate hardcoded names with those passed down in props.
e.g. `className={cx('scroll', props.className)}`

ciaranschutte: this isn't actually in use, didn't catch that to remove. all styling done through the `css` prop

components/pages/discovery/components/api/Chart.tsx
ciaranschutte: https://github.com/icgc-argo/platform-ui/pull/2795/commits/372df3fd59bc98a39d20a26ee4a559250f006574

justincorrigible (approved):

PR #2801, opened 2025-02-12
Charts

ciaranschutte:
# Description of changes

- Adds Arranger targeted charts lib `charts/..`
  - Bar
  - Line
  - Doughnut (custom chart UI, not direct Nivo import, **_this is currently a skeleton until data hookup clarified_**)
  - Uses apollo client to query Arranger, does not use Arranger v3
- Implements ARGO charts using lib, passing in additional "consumer config" configuration for styling and customization

example use: https://github.com/icgc-argo/platform-ui/pull/2801/files#diff-585e6deecbc61bbc1fd2fdd94286e66f32f1278f2a962d9a2f68820a49db45beR92

#### UI Consumer

Responsibility: Styling and additional customization of chart level
examples:
- `<Bar field='study_id' consumerConfig={...barChartStyles}/>`

#### Arranger Chart Lib
Responsibilities:
- Data querying & formatting for chart (internal)
- Render chart with provided consumer config

examples:
- query Arranger server for field, format data for chart, render chart
- call `onError` callback provided by end user if query fails (important for end user UI)


charts/Chart.tsx
```diff
@@ -0,0 +1,62 @@
...
+        onError && onError(error);
+      }
+    }, [loading, error]);
+
+    if (!loading && !error) {
+      // provides resolved data to user config function, assist in UI styling
+      const resolvedConsumerConfig =
+        typeof consumerConfig === 'function' ? consumerConfig({ data }) : consumerConfig;
+
+      return <Component {...{ data, ...internalConfig, ...resolvedConsumerConfig }} />;
```
ciaranschutte, line 56: nested spread to make it clear that configs will override each other.
alternative is:
```react
 <Component {...internalConfig} {...resolvedConsumerConfig } />
 ```

File: charts/Bar.tsx [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+    );
+  };
+
+const Bar = (consumerProps) => {
+  const { field } = consumerProps;
+  return generateChartComponent({
+    Component: ResponsiveBar,
+    options: { query: generateQuery({ field }), dataTransformer: transformToBarData({ field }) },
+    internalConfig: { ...BUCKETS_TO_BAR_CHART },
+  })(consumerProps);
```
ciaranschutte, line 57: use function to generate chart component. clean, minimal interface here, nothing too complicated to follow for a higher order func.

alternative is a second React component but there's nothing other than a render just to pass props. not needed.

charts/arranger.ts
```diff
@@ -0,0 +1,67 @@
...
+`;
+
+export const useArrangerCharts = ({ options: { query, variables, dataTransformer } }) => {
+  console.log('data tran', query, variables, dataTransformer);
+  const { filters } = useFiltersContext();
+  const {
+    data: rawData,
+    loading,
+    error,
+  } = useQuery(query, { variables: { ...variables, filters } });
```
ciaranschutte, line 56: this entire file will be a rewrite for Arranger v3
follows our ARGO code for now. uses apollo `useQuery`.

it does mean every chart will have a query which is hopefully working correctly with the ARGO apollo client, it does look like there are a lot of queries when we should be using one. (gql - ask for what you want)

charts/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import Doughnut from './Doughnut';
+import Line from './Line';
+
+const Charts = {
+  Bar,
+  Line,
+  Doughnut,
+};
+
+export default Charts;
```
ciaranschutte, line 30: barrel file

charts/types.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+// TODO: chart configs are different for differnet charts
+export type ChartConfig = {};
+
+export type Chart = {
+  consumerConfig?: ChartConfig;
+  onLoad?: () => void;
+  onError?: (error: unknown) => void;
+};
```
ciaranschutte, line 27: will be expanding this in future PR

components/pages/discovery/index.tsx
```diff
@@ -17,18 +17,18 @@
...
 import Head from '../head';
 
-import Charts from './Charts';
-import StatsCard from './components/StatsCard';
-import QueryBarContainer from '../file-repository/QueryBar/QueryBarContainer';
-import Sidebar from './components/SideBar';
 import Footer from 'components/Footer';
 import { useState } from 'react';
+import QueryBarContainer from '../file-repository/QueryBar/QueryBarContainer';
+import Charts from './Charts';
```
justincorrigible, line 28: at first glance, it is not 100% straightforward whether this is `Charts.tsx` or a possible `charts/index.tsx` as we're not using ESM in this project. case sensitivity aside, this can be ambiguous...
I'd rename Charts.tsx to something like chartsLayout or better yet, move it to be the index file in the charts folder? (and thus keep the existing barrel pattern)

components/pages/discovery/charts/Line.tsx
```diff
@@ -0,0 +1,70 @@
...
+        text: {
+          fontSize: 6,
+        },
+      },
+    },
+  };
+
+  return (
+    <Chart>
+      <Charts.Line fields={fields} interval={interval} consumerConfig={config} />
```
justincorrigible, line 65: when compared to `Charts.whatever` being a whatever chart, I find unclear what the Chart parent component does or why it's needed. is it a data/theming context?
the name is not descriptive of its purpose

charts/Chart.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ({
+    // internal
+    Component,
+    options,
+    internalConfig,
+  }: {
+    Component: ChartComponent;
+    options: Options;
+    internalConfig?: Record<string, unknown>;
+  }) =>
```
joneubank, line 57: Using inferred generic type for the component so we get type information on the `internalConfig` params.
```suggestion
const generateChartComponent =
  <TComponent extends ChartComponent>({
    // internal
    Component,
    options,
    internalConfig,
  }: {
    Component: TComponent;
    options: Options;
    internalConfig?: Omit<ComponentProps<TComponent>, 'data'>;
  }) =>
```

components/pages/discovery/Charts.tsx
```diff
@@ -276,7 +107,7 @@ const Charts = () => {
           gridRowEnd: 3,
         })}
       >
-        <DoughnutChart data={doughnutChartData} config={{}} />
+        <DoughnutChart field={undefined} />
```
joneubank, line 110: Assume this is a big `// TODO`

components/pages/discovery/Charts.tsx
```diff
@@ -288,20 +119,20 @@ const Charts = () => {
           gridRowEnd: 5,
         })}
       >
-        <BarChart data={donorData} config={programIDBarChart} />
+        <BarChart field="study_id" />
```
joneubank, line 122: I love how straightforward this is to use.

charts/Chart.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export type Options = {
+  query: DocumentNode;
+  variables?: Record<string, unknown>;
+  dataTransformer?: (data: unknown) => any;
+};
+
+/**
+ *
+ * Higher order function to return a chart component with a standard end user api and data resolution
```
justincorrigible, line 33: found it! this is Chart, and it seems to be doing the same job as a Charts data context, without actually being a context provider... 🤔 curious to see what the effect could be when there's state changes, or its performance impacts (if any)

justincorrigible (approved):
Lots of good work... save some comments I left about confusing naming choices, this should be okay to merge as is.

joneubank (changes requested):
One suggestion to improve types.

There is one question I have that I am struggling to answer... where are you telling the charts library what the host endpoint is for arranger? This is a parameter that will need to be passed into the charts component library, but I couldn't find this parameter... so I assume youre just importing the argo config driectly into your component library. Maybe you could point it out for me.

charts/Chart.tsx
ciaranschutte: Nice! This is why I added the `ChartComponent` type but React isn't happy with it when I tried to pass a generic. `React.ComponentType` should work but there's some clashes on prop types.

joneubank:
The apollo provider config is sufficient for now but we'll need to abstract the apollo client into the charts lib to make it reusable, maybe the charts should accept the arranger apollo client as a prop? Or something like that.

PR #2819, opened 2025-07-21
Feature/data discovery page feedback

ciaranschutte:
# Description of changes

## Type of Change

- [ ] Bug
- [ ] Dependency updates
- [ ] Feature
- [ ] Refactoring
- [ ] Release candidate
- [ ] Styling
- [ ] Testing
- [ ] Other (please describe)

- Design (select one):
  - [ ] Matches design:
    - font size, weight, colour
  - [ ] No changes to UI
- Back-end (select one):
  - [ ] No back-end changes
- [ ] Manually tested changes
- [ ] Connected ticket to PR

### DIFF
--- a/charts/Doughnut/ui.tsx
+++ b/charts/Doughnut/ui.tsx
@@ -88,6 +88,12 @@ export const DoughnutChart = ({ data, config }) => {
         flexDirection: 'row',
         width: '100%',
         height: '100%',
+        pointerEvents: 'none',
+
+        // prevent overlapping of elements from obstructing "path:hover"
+        path: {
+          pointerEvents: 'auto',
+        },
       })}
     >
       <div css={css({ height: '100%', width: '70%', position: 'relative' })}>
--- a/components/pages/discovery/Charts.tsx
+++ b/components/pages/discovery/Charts.tsx
@@ -154,6 +154,10 @@ const ChartsLayout = () => {
               // @ts-expect-error slight difference in specificity between writing a direct SQON filter and unofficial FileRepo types
               setSQON(toArrangerV3Filter(sqonFilter));
             },
+            onDataLoad: (data) => {
+              // order data, range query so there won't be "no data"
+              return data.toReversed();
+            },
           }}
         />
       </Card>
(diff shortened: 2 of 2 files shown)


charts/Doughnut/ui.tsx
```diff
@@ -35,9 +35,12 @@ const Legend = ({ data }: { data: { label: string; color: string }[] }) => {
         '> div': { marginTop: '16px' },
       })}
     >
-      {data.map((legend) => {
+      {data.map((legend, index) => {
         return (
-          <div css={css({ display: 'flex', flexDirection: 'row', alignItems: 'center' })}>
+          <div
+            css={css({ display: 'flex', flexDirection: 'row', alignItems: 'center' })}
+            key={index}
```
joneubank, line 42: Is there a better key you can use than the index? https://react.dev/learn/rendering-lists#why-does-react-need-keys

components/pages/discovery/Charts.tsx
```diff
@@ -182,40 +185,39 @@ const ChartsLayout = () => {
...
         <Barchart
           fieldName="gender"
           theme={{
+            ...commonTheme,
             onClick: (config) => chartFilters.gender(config.data.key),
-            axisLeft: { legend: 'Genders' },
-            axisBottom: { legend: 'Donors' },
+            onDataLoad: (data) => {
+              return data.toReversed();
+            },
```
joneubank, line 201: This is purely stylistic and pointing out an inconsistency in style - why on this function specifically don't we use an implicit return, while for the `onClick` function just above it we do:

```ts
onDataLoad: (data) => data.toReversed(),
```

components/pages/discovery/components/Facets/index.tsx
```diff
@@ -68,17 +64,24 @@ const FacetCollection = ({
...
+      {staticFacets.map(({ name, contents }, index) => {
         return (
           <FacetFolder
             title={name}
             onClick={() =>
               setVisiblePanels({ type: FACET_VISIBILITY_TOGGLE_ACTIONS.TOGGLE_FOLDER, name })
             }
             isExpanded={isFolderExpanded(name)}
-            key={`name_${idx}`}
+            key={`name_${index}`}
```
joneubank, line 75: Using index as key but with prefix... I don't think the prefix does anything here. Is there a better value to use as the key?

joneubank (approved):
Lots of good little things, thanks.

components/pages/discovery/Charts.tsx
ciaranschutte: nice catch - preference for explicit - will adjust

PR #2840, opened 2025-09-02
Feature/data discovery qa feedback

ciaranschutte:
- use latest @overture-stack/arranger-charts dependency
- remove platform-ui redundant charts
- update stats card hookup
- use sunburst mapping function instead of object
- show "max data" on charts card 
- styling, syntax fixes


components/pages/discovery/components/VisibleElements.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const VisibleElements = ({ maxElements, fieldName }) => {
+  const { getChartData } = useChartsContext();
+  const { isLoading, isError, data } = getChartData(fieldName);
+
+  const noData = isLoading || isError || (data && data?.length === 0);
+  const totalWithinMax = !noData && data.length <= maxElements;
+
+  return noData || totalWithinMax ? null : (
+    <div>{`Top ${Math.min(maxElements, data.length)} of ${data.length}`}</div>
+  );
```
joneubank, line 40: This probably works, but is a tough read - I think because it uses negative logic. Maybe it simplifies with:


```suggestion
  const hasData = !isLoading && !isError && data && data.length > 0;
  const hasMoreThanMaxElements = hasData && data.length >= maxElements;

  return hasMoreThanMaxElements ?  (
    <div>{`Top ${Math.min(maxElements, data.length)} of ${data.length}`}</div>
  ) : null;
```

joneubank (approved):

PR #2846, opened 2025-09-10
use latest arranger charts

ciaranschutte:
- use latest arranger charts

### DIFF
--- a/components/pages/discovery/Charts.tsx
+++ b/components/pages/discovery/Charts.tsx
@@ -129,15 +129,6 @@ const ChartContainer = ({ children }) => (
   </div>
 );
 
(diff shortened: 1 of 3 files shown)


components/pages/discovery/Charts.tsx
```diff
@@ -169,7 +160,7 @@ const ChartsLayout = () => {
     primary_site: chartFilter('primary_site'),
     vital_status: chartFilter('vital_status'),
     analyses__experiment__experimental_strategy: chartFilter(
-      'analyses.experiment.experimental_strategy',
+      'analyses.exper`iment.experimental_strategy',
```
justincorrigible, line 163: is this correct?

PR #2852, opened 2025-09-30
Feature/announcement

ciaranschutte:
- two announcements
- - data discovery page
- - data release 13


justincorrigible:
nice catch with the branch. `main` was mindlessly default for me
🤦 argo

